Issue #7, opened 2022-09-07
Adding an option to pass `--extra-arg` to clang-tidy


hillmich:
For a header-only library I'm maintaining the clang-tidy checks fail with warnings about c++17 features being used. This is most likely cause by clang-tidy assuming C++14 for the header files (since it's the default). Running `clang-tidy` manually with ` --extra-arg="-std=c++17"` fixes (most) problems. You can the library at cda-tum/dd_package#104

Could you add a parameter to `cpp-linter` to pass along extra arguments to `clang-tidy`?

2bndy5:
sounds reasonable. Does your generated database not have the `-std` arg in it?

hillmich:
The generated database has the `-std` flag, but it seems clang-tidy doesn't pick it up automatically for the headers.

Just for completeness, the database for the source files looks like this

```
cat build/compile_commands.json
[
{
  "directory": "/home/runner/work/dd_package/dd_package/build/test",
  "command": "/usr/bin/clang++-14  -I/home/runner/work/dd_package/dd_package/include -O3 -DNDEBUG -Wall -Wextra -pedantic -fno-math-errno -ffinite-math-only -fno-trapping-math -mtune=native -march=native -std=c++17 -o CMakeFiles/DDPackage_example.dir/main.cpp.o -c /home/runner/work/dd_package/dd_package/test/main.cpp",
  "file": "/home/runner/work/dd_package/dd_package/test/main.cpp"
}, ...
```

PR #8, opened 2022-09-07
Added extra parameter to pass extra arg to clang-tidy

hillmich:
Solves cpp-linter/cpp-linter#7 by adding a parameter `--extra-arg` to `cpp-linter`.

### DIFF
--- a/cpp_linter/run.py
+++ b/cpp_linter/run.py
@@ -237,6 +237,13 @@ file annotations as feedback.
 
 Defaults to ``%(default)s``.""",
 )
+cli_arg_parser.add_argument(
+    "--extra-arg",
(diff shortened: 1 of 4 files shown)


2bndy5:
I was just about to publish my patch 😃 ... It seems you did almost everything I did, but I added to the docs a bit more (and added `-x` as a short CLI alternative).

Would you be offended if I close this and PR my branch instead?

hillmich:
> Would you be offended if I close this and PR my branch instead? [main...tidy-extra-args](https://github.com/cpp-linter/cpp-linter/compare/main...tidy-extra-args)

Not at all, please go ahead.

PR #9, opened 2022-09-07
addresses #7

2bndy5:


2bndy5:
@hillmich You could test the patch in your CI by using
```
pipx run --spec git+https://github.com/cpp-linter/cpp-linter@tidy-extra-args cpp-linter
```

cpp_linter/run.py
```diff
@@ -511,6 +524,8 @@ def run_clang_tidy(
         line_ranges = dict(name=filename, lines=file_obj["line_filter"][ranges])
         logger.info("line_filter = %s", json.dumps([line_ranges]))
         cmds.append(f"--line-filter={json.dumps([line_ranges])}")
+    if extra_args:
+        cmds.append(f"--extra-arg={repr(extra_args)}")
```
hillmich, line 528: This fails in my test (installed `cpp-linter` from your branch) with the `repr(...)`. 

```
$ cpp-linter --version=12 --extra-arg=-std=c++17 ...
...
INFO:CPP Linter:Running "clang-tidy-12 --export-fixes=clang_tidy_output.yml -p /home/stefan/repos/quantum/dd_package/build --extra-arg='-std=c++17' test/bench_package.cpp"
DEBUG:CPP Linter:Output from clang-tidy:
error: no such file or directory: ''-std=c++17''; did you mean '-std=c++17''? [clang-diagnostic-error]
```

Notice the extra `'` around `-std=c++17`.

hillmich: Removing the `repr(...)` call fixes the problem in my test and the checks go through smoothly.

2bndy5: Its best to use quotes around argument values that resemble CLI options. Otherwise you risk confusing the `argparse` python lib (& in this case clang-tidy). I'll have to add this to the docs as well.

hillmich: [On the Github CI](https://github.com/cda-tum/dd_package/runs/8228327913?check_suite_focus=true), you can see error as well.

In this case I think there is only a low risk of confusing clang-tidy as we explicitly pass the whole string as *one* parameter (it's one item in the list and there is no additional processing by a shell or similar). Wenn I ran it locally without the `repr`, `argparse` seemed fine as well.

2bndy5: The problem I'm trying to avoid is users passing more than 1 arg to `extra-args`.
```
cpp-linter -x=-std=c++14 -Wall
```
But it would seem as though clang-tidy expects only 1 arg and allows the `extra-arg` to be specified more than once. Python's `argparse` has a way of handling this...

2bndy5: implementing this in the composite action (cpp-linter-action repo) will be difficult because they don't allow a variety of options as accepted forms of action inputs (pretty much string or bool).

hillmich: Having the argument `--extra-args` (notice the `s`) and then split on whitespace would be an option. Of course, this fails if the extra argument has a space or similar. Do you think this could be an acceptable solution?

2bndy5: I prefer to have `argparse` do the parsing for me:
```py
cli_arg_parser.add_argument(
    "-x",
    "--extra-arg",
    default=[],
    action="append",
```
But your idea is the approach I'd have to take for the composite action repo.

hillmich: On second thought, using [`action='append'`](https://docs.python.org/3/library/argparse.html#action) in the argument parser seems like a better solution.

2bndy5: This might be more confusing, but I guess we could accept both `--extra-args` and `--extra-arg` (the later numerous times) and join them together...

> Of course, this fails if the extra argument has a space or similar

I don't immediately see a solution for this. I understand this is the typical way to pass an arg in CLI on Linux, but Windows would interpret the following as 2 separate args
```
cpp-linter -x "-std=c++14"
```
It might be best to strongly encourage delimiting arg-name/arg-value with `=`.

hillmich: In my opinion having both ways to specify extra arguments is more confusing. Using `action='append'` would be in line with how clang-tidy works and avoids us having to be overly clever. But I'm fine with both approaches :)

That windows parses differently is unfortunate, but I think this can be "fixed" with appropriate documentation.

2bndy5: Good feedback!

hillmich:
The changes look good to me and they work locally and in the CI of my project :+1:

shenxianpeng (approved):
Sorry for the wait, it looks good to me.

cpp_linter/run.py
```diff
@@ -489,6 +507,23 @@ def run_clang_tidy(
...
+        arguments.
+
+        .. note::
+            If the list is only 1 item long and there is a space in the first item,
+            then the list is reformed from splitting the first item by whitespace
+            characters.
+
+            .. code-block:: shell
+
+                cpp-linter -extra-arg="-std=c++14 -Wall"
```
shenxianpeng, line 520:
```suggestion
                cpp-linter --extra-arg="-std=c++14 -Wall"
```

shenxianpeng: The example code seems miss a `-`

shenxianpeng: Run `cpp-linter --help`, and the output contains the syntax of the document, like `.. code-block:: shell`, I guess it works as designed

```bash
-x EXTRA_ARG, --extra-arg EXTRA_ARG
                        A string of extra arguments passed to clang-tidy for use as
                        compiler arguments. This can be specified more than once for each
                        additional argument. Recommend using quotes around the value and
                        avoid using spaces between name and value (use ``=`` instead):
                        
                        .. code-block:: shell
                        
                            cpp-linter --extra-arg="-std=c++17" --extra-arg="-Wall"
                        
                        Defaults to ``'[]'``.
```

2bndy5: Oh I missed this comment.

I'll add it to main.

2bndy5: Fixed. Good eye!

PR #13, opened 2022-10-21
Remove set-output instead of GITHUB_OUTPUT

shenxianpeng:


2bndy5:
This should be tested in cpp-linter-action before merging. The unit tests here only verify the exit code returned by the function. https://github.com/cpp-linter/cpp-linter/blob/a23e14ddcfaef83da0f4aca740dbfd547163e207/tests/test_misc.py#L20-L22

I guess we could also test that the stdout contains the necessary GitHub workflow command.

cpp_linter/run.py
```diff
@@ -266,7 +266,7 @@ def set_exit_code(override: int = None) -> int:
         then this value will describe (like a bool value) if any checks failed.
     """
     exit_code = override if override is not None else bool(Globals.OUTPUT)
-    print(f"::set-output name=checks-failed::{exit_code}")
+    os.environ["GITHUB_OUTPUT"] = f"checks-failed={exit_code}"
```
2bndy5, line 269: I'm not sure if this will work with GitHub's runner context. [The GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) say we should write/append the var (as a `key=value`) to a file pointed to by `GITHUB_ENV`.

2bndy5: I'm trying to say that the step's env is flushed when the step finishes. That's why github recommends writing to a "env file" so the env change can persist across steps.
```python
    try:
        with open(os.environ["GITHUB_OUTPUT"], "a", encooding="utf-8")as env_file:
            env_file.write(f"checks-failed={exit_code}")  # maybe add `\n` if using multiple vars?
    except FileNotFoundError:
        # not executed on a github CI runner; ignore this error when executed locally
        pass
```

2bndy5: The examples from that page are still `echo`ing to a file (`>>`) - not stdout.

cpp_linter/run.py
```diff
@@ -266,7 +266,12 @@ def set_exit_code(override: int = None) -> int:
         then this value will describe (like a bool value) if any checks failed.
     """
     exit_code = override if override is not None else bool(Globals.OUTPUT)
-    print(f"::set-output name=checks-failed::{exit_code}")
+    try:
+        with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8")as env_file:
```
2bndy5, line 270:
```suggestion
        with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8") as env_file:
```

tests/test_misc.py
```diff
@@ -17,9 +18,12 @@
 )
 
 
-def test_exit_override():
+def test_exit_override(tmpdir):
```
2bndy5, line 21: [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures) recommend using the newer fixture `tmp_path`, which is like `tmpdir` but uses `pathlib.Path`.

According to the [`py.path` docs](https://py.readthedocs.io/en/latest/path.html#py-path), it in maintenance mode and not recommended for newer projects. So, this should really be changed to `tmp_path` fixture.

2bndy5 (changes requested):

2bndy5:
I have the IDE open now. I'm going to push the requested changes and then approve this, ok?

2bndy5 (approved):

2bndy5:
Looks like there was a connection error in uploading the coverage report (can be fixed by re-running the failed job), otherwise this is passing in CI.

2bndy5:
I think I screw up codecov.io when I do force pushes. But that's really on them because they need a better way of analyzing the git history...

PR #15, opened 2022-11-05
add code style "chromium", "microsoft" to file annotation

### LINKED ISSUE #14: Support coding style Microsoft to file annotation
Before clang-format 10 only supported following code style by default
LLVM, Google, Chromium, Mozilla, WebKit.

Since clang-format 11 supported GNU and Microsoft by default.
LLVM, GNU, Google, Chromium, Microsoft, Mozilla, WebKit.

How about supporting GNU and Microsoft to cpp-linter?

shenxianpeng:
Fix #14

also, adjust the order a bit to make it look consistent with the order of the code style output from `clang-format --help` for easy comparison.

```
 --style=<string>               - Coding style, currently supports:
                                     LLVM, GNU, Google, Chromium, Microsoft, Mozilla, WebKit.
```

### DIFF
--- a/.github/workflows/pre-commit-hooks.yml
+++ b/.github/workflows/pre-commit-hooks.yml
@@ -12,6 +12,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.8'
+          python-version: '3.10'
       - run: python3 -m pip install pre-commit
       - run: pre-commit run --all-files


.github/workflows/pre-commit-hooks.yml
```diff
@@ -12,6 +12,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.x'
+          python-version: '3.8'
```
2bndy5, line 15: Please undo this. I want to be alerted of any problems with the latest version of python, even if that is not what I use locally,

If this has something to do with the specific version required by clang tools, then it should be addressed in a CI that actually uses the clang tools.

2bndy5 (changes requested):

2bndy5:
> run `pre-commit run --all-files` successfully from my environment, not sure why it failed in the workflow.

I think the pylint settings just need to be updated. I see this in the CI log:
> ************* Module /home/runner/work/cpp-linter/cpp-linter/pyproject.toml
> pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'missing-timeout' (unknown-option-value)

This message usually indicates that pylint was updated and the update brought changes that encountered an unexpected config option.

.github/workflows/pre-commit-hooks.yml
2bndy5: Well, python v3.11 was released today. I don't see anything in the [3.11 changelog](https://docs.python.org/3/whatsnew/3.11.html) related to the errors the failed CI logs. I also checked with the [pylint changelog](https://github.com/PyCQA/pylint/releases) and I don't see anything related to removing the `missing-timeout` check.

I suppose we can avoid testing with python v3.11 or newer while they work out the "day 1" bugs.

2bndy5, line 15:
```suggestion
          python-version: '3.10'
```
This is the version I have installed locally, and it works fine.

2bndy5 (approved):

PR #19, opened 2022-11-08
Error: Codecov: Failed to properly upload

shenxianpeng:
More about this problem can refer to https://github.com/codecov/codecov-action/issues/598

### DIFF
--- a/.github/workflows/run-dev-tests.yml
+++ b/.github/workflows/run-dev-tests.yml
@@ -109,7 +109,7 @@ jobs:
       - uses: codecov/codecov-action@v3
         if: matrix.os == 'ubuntu-latest' && matrix.version == '12' && matrix.py == '3.10'
         with:
-          token: ${{secrets.CODECOV_UPLOAD_TOKEN}}
+          token: ${{secrets.CODECOV_TOKEN}}
           files: ./coverage.xml
           fail_ci_if_error: true # optional (default = false)
           verbose: true # optional (default = false)


.github/workflows/run-dev-tests.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       - uses: codecov/codecov-action@v3
         if: matrix.os == 'ubuntu-latest' && matrix.version == '12' && matrix.py == '3.10'
         with:
+          token: ${{secrets.CODECOV_UPLOAD_TOKEN}}
```
2bndy5, line 112: This shouldn't be needed if the secret is named `CODECOV_TOKEN`. Notice I added such a secret to clang-tools-pip repo, and the codecov action was able to use that automatically.

2bndy5: I think the `CODECOV_TOKEN` just needed to be regenerated and updated.

> another is an analysis token

Not sure what you mean here. We only use the `CODECOV_TOKEN` to upload reports to codecov.io

2bndy5: image
Still needs an updated value I think

2bndy5: > It is called Impact analysis token

Oh! I didn't know about that feature. It is still in beta now anyway. I think our reports already include contextual information about what test function specifically covers certain lines, but codecov doesn't use it. Maybe this beta feature is an attempt to make use of it.

BTW, if you run
```
coverage run -m pytest
coverage html
```
then the generated html output will tell you what test function used the covered line:
image

2bndy5 (approved):

Issue #18, opened 2022-11-03
lines-changed-only gets confused when lines only deleted from a file


turol:

2bndy5:
@shenxianpeng To be clear, the desired behavior here is to skip the file entirely if there are now lines added. The solution you propose in #21 doesn't satisfy that behavior.

PR #21, opened 2022-11-12
Skip file when no new lines added and lines-changed-only is asserted

shenxianpeng:
I think the current change can stop adding `--line-filter` arg in clang-tidy when `lines` is empty. 

But to skip adding annotation for such situation (only delete code), I guess need to change func `range_of_changed_lines` to get an empty list, but I am not sure how to change it. @2bndy5 could you help when you have time?

https://github.com/cpp-linter/cpp-linter/blob/f53976ac55b114584dab3bf2a49a3da12075fc25/cpp_linter/run.py#L890

I used this [test job](https://github.com/shenxianpeng/test-repo/actions/runs/3450958298/jobs/5759870745) to reproduce #18

### DIFF
--- a/cpp_linter/__init__.py
+++ b/cpp_linter/__init__.py
@@ -4,7 +4,7 @@ import os
 from pathlib import Path
 import platform
 import logging
-from typing import TYPE_CHECKING, List, Dict, Tuple, Any
+from typing import TYPE_CHECKING, List, Dict, Tuple, Any, Union
 from requests import Response
 
 if TYPE_CHECKING:  # Used to avoid circular imports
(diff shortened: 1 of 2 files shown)


2bndy5:
The solution for this is going to need more than just a few lines changed. I'll start hacking it...

2bndy5:
This branch is now overwritten by my changes to properly address #21. Coverage report seems to have suffered (for some unforeseen reason), but most of thread_comments.py is not actually used in production.

2bndy5:
The unit tests agree with you. If released, this would be a patch version bump. I'm ready when you are.

2bndy5:
FYI, I added your p_enemy.c file to the unit test for precision.

2bndy5:
> So I need to keep p_enemy.c file always in the test-repo? at least I need to keep test-repo always existing, I think I will not delete it.

You shouldn't need to since the file's "added" state is specific to the commit that was pushed to the remote. I don't think GitHub deletes orphaned branches/commits.

2bndy5:
We could always change the file to a new file in this repo's demo src. Ironically, the name p_enemy.c is also used in chocolate-doom (which we also use in the unit test), but there's no conflict because chocolate-doom's file is located in src/hexen/p_enemy.c (your file is in the supposed repo-root).

PR #33, opened 2023-05-20
Sonar cloud concerns

2bndy5:
This is mostly code smell problems that are easy to fix.

As for the regex security hotspot, I tried to make the pattern more specific, but the nature of the string being matched makes this nearly impossible to address.

### DIFF
--- a/cpp_linter/clang_tidy.py
+++ b/cpp_linter/clang_tidy.py
@@ -5,7 +5,7 @@ import re
 from typing import Tuple, Union, List, cast
 from . import GlobalParser, CLANG_TIDY_STDOUT
 
-NOTE_HEADER = re.compile(r"^([^:]+):(\d+):(\d+):\s(\w+):(.*)\[([a-z\d\-]+)\]$")
+NOTE_HEADER = re.compile(r"^(.+):(\d+):(\d+):\s(\w+):(.*)\[([a-zA-Z\d\-\.]+)\]$")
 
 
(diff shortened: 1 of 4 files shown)


cpp_linter/clang_tidy.py
```diff
@@ -5,7 +5,7 @@
 from typing import Tuple, Union, List, cast
 from . import GlobalParser, CLANG_TIDY_STDOUT
 
-NOTE_HEADER = re.compile(r"^(.+):(\d+):(\d+):\s(\w+):(.*)\[(.*)\]$")
+NOTE_HEADER = re.compile(r"^([^:]+):(\d+):(\d+):\s(\w+):(.*)\[([a-z\d\-]+)\]$")
```
2bndy5, line 8: My main concern here with `.+` ➡️ `[^:]+` is that Linux actually allows colons (`:`) in the file path/name. Windows and MacOS don't allow colons to be used in file names/paths (probably a legacy decision due to transport protocols like `file:///`). However, we can't assume that all users projects will be cross-platform friendly.

I also have to double check the numerous names of clang-tidy diagnostics to make sure `[a-z\d\-]+` is a compatible change.

2bndy5: I have reverted the start of this pattern back to `.+` for cross-platform compatibility.

I also adjusted the pattern that matches the clang-tidy diagnostic name.

PR #35, opened 2023-09-02
resolve #34

### LINKED ISSUE #34: New inputs to provide finite control of `thread-comments`
See https://github.com/cpp-linter/cpp-linter-action/issues/158#issuecomment-1690106555

- [x] I think we can augment the `thread-comments` input, where
   - `true` would be delete and post anew (the current behavior)
   - `false` disable feature (as has always been)
   - `update` would be to re-use existing comment or create a new if none exists (this is the new suggestion).

- [x] A `no-lgtm` option would be needed to prevent the new "LGTM" comment from getting posted (as was the behavior before v1.6.0). I have to look at the history, but I think we also used to leave an outdated comment in the thread when checks passed. We should delete our previous PR comment if checks passed and an old one exists. @shenxianpeng Do you object to making this `no-lgtm` option default to true?

---
If I have the time, this would also be a good opportunity to make sure we check the full list of PR comments instead of only the first 30 (or whatever the REST API payload limit is).

2bndy5:
This still needs testing.

Also updated the pre-commit hooks. I replaced pylint with [ruff](https://beta.ruff.rs/) for better speed and because we already get type-checking from mypy; we don't need the extra type-checking from pylint.

Removed _some_ dead code from run.py. I have some other optimizations I'd like to add from the deep-dive code review I did when porting this pkg to dart. But that can happen in another PR.

I found and fixed a couple bugs in `make_annotations()`:
1. If `file-annotations` is false, then clang-tidy annotations were still posted.
2. If `file-annotations` is false, then clang-format advice was not counted.

I also modified the log output to show the count of _possible_ annotations which directly corresponds to the `checks-failed` output.


2bndy5:
Codecov CI is complaining about lines that changed in this PR but are not covered with unit tests.

Honestly, we don't have an adequate way to test anything involving Github's REST API usage (not without hitting the REST API's usage rate limit). So, the codecov complaints are null because the functionality wasn't unit tested before anyway.

2bndy5:
The overall coverage  went up, but the diff coverage is below some target that codecov seems to have set (81.48%).

2bndy5:
So far, I've tested the following configs:
```yml
        with:
          files-changed-only: true | false
          thread-comments: update
          no-lgtm: false
```
```yml
        with:
          files-changed-only: true | false
          thread-comments: true
          no-lgtm: false
```
Now, I have to test the same `thread-comments` inputs with `no-lgtm: true`.

2bndy5:
Found and fixed a couple bugs in `make_annotations()`:
1. If `file-annotations` is false, then clang-tidy annotations were still posted.
2. If `file-annotations` is false, then clang-format advice was not counted.

I also modified the log output to show the count of _possible_ annotations which directly corresponds to the `checks-failed` output.

2bndy5:
Code coverage is missing some lines in `make_annotations()` now. I'd like to add a unit test to get those lines tested properly. Other lines in these changes can't be covered without using the GitHub REST API, so I won't worry about any of that as long as everything checks out on the test repo.

2bndy5:
> Code coverage is missing some lines in make_annotations() now. I'd like to add a unit test to get those lines tested properly. 

TL;DR I need to find a better commit for the `test_tidy_annotations()`

After playing around with `test_tidy_annotations()` (which should cover these changes), I found that the commit we're using doesn't generate tidy advice for every condition (all values of `lines_changed_only` and `tidy_checks` using either config file or action defaults).

2bndy5:
I also found that the filename matching was flawed in `make_annotations()`. So, I fixed it and updated the unit test... Should be done now.

cpp_linter/run.py
```diff
@@ -678,30 +589,34 @@ def make_annotations(
...
             for file in files:
+                print(filename, "?=", file["filename"])
                 if filename == file["filename"]:
                     line_filter = cast(
                         List[int], range_of_changed_lines(file, lines_changed_only)
                     )
                     break
-            else:
+            else: # filename match not found; treat line_filter as empty list
                 continue
```
2bndy5, line 610: I _think_ we should error here if filename was not matched, but I can't remember why I had previously set it to `continue`. If no filename match was found then, clang-tidy notes are not filtered by line changes, which seems ok for fallback behavior.

2bndy5: Maybe if filename match wasn't found then it is possibly from a file in C/C++ std libs. Probably better to leave as is.

2bndy5: It is not a syntax error; see [grammar of python `for` loop](https://docs.python.org/3/reference/compound_stmts.html#the-for-statement). The `else` block is only triggered if the `for` loop never `break`s:
```python
for i in range(3):
    if i > 3:
        break
else:
    # usually this is where errors are dealt with
    print("`i` was never greater than 3")
```

2bndy5:
Everything passed on the second round of tests in cpp-linter/test-cpp-linter-action#22. I'm now confident enough to merge this! @shenxianpeng Thanks for your patience.

PR #36, opened 2023-10-30
fix format comment

2bndy5:
in reference to cpp-linter/cpp-linter-action#161

May need to adjust unit tests for this

### DIFF
--- a/cpp_linter/run.py
+++ b/cpp_linter/run.py
@@ -418,15 +418,14 @@ def create_comment_body(
     if CLANG_FORMAT_XML.exists() and CLANG_FORMAT_XML.stat().st_size:
         parse_format_replacements_xml(PurePath(filename).as_posix())
         if GlobalParser.format_advice and GlobalParser.format_advice[-1].replaced_lines:
-            should_comment = lines_changed_only == 0
-            if not should_comment:
-                for line in [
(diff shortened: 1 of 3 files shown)


2bndy5:
Code coverage decrease is because we aren't uploading reports for unit tests on Windows. It is not a direct result of these changes. These changes are all covered by current unit tests. I admit, the unit test covering these lines should be more exhaustive to detect the reported bug.

cpp_linter/run.py
```diff
@@ -478,9 +478,10 @@ def capture_clang_tools_output(
     if Globals.FORMAT_COMMENT or Globals.TIDY_COMMENT:
         Globals.OUTPUT += ":warning:\nSome files did not pass the configured checks!\n"
         if Globals.FORMAT_COMMENT:
+            files_list = re.findall(r"\- [^\n]\n", Globals.FORMAT_COMMENT, re.MULTILINE)
```
2bndy5, line 481:
```suggestion
            files_list = re.findall(r"\- [^\n]+\n", Globals.FORMAT_COMMENT, re.MULTILINE)
```

2bndy5:
I found a more efficient way to get the count of files that need formatting (doesn't use regex pattern anymore).

PR #37, opened 2023-12-02
add more specific outputs

2bndy5:
satisfies feature requested in cpp-linter/cpp-linter-action#168

### DIFF
--- a/cpp_linter/run.py
+++ b/cpp_linter/run.py
@@ -61,10 +61,18 @@ def set_exit_code(override: Optional[int] = None) -> int:
         The exit code that was used. If the ``override`` parameter was not passed,
         then this value will describe (like a bool value) if any checks failed.
     """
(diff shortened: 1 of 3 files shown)


cpp_linter/run.py
```diff
@@ -61,10 +61,18 @@ def set_exit_code(override: Optional[int] = None) -> int:
         The exit code that was used. If the ``override`` parameter was not passed,
         then this value will describe (like a bool value) if any checks failed.
     """
-    exit_code = override if override is not None else bool(Globals.OUTPUT)
+    exit_code = (
+        override
+        if override is not None
+        else (Globals.format_failed_count + Globals.tidy_failed_count)
+    )
```
2bndy5, line 68: Using `bool(Globals.OUTPUT)` always returns true since revision of comment contents (see changes in #32). So, this enhancement fixes that ambiguity while implementing the new outputs. 🚀

PR #40, opened 2023-12-13
use pygit2 and improve file info handling

2bndy5:
- Increased code coverage
- Faster execution by using `pygit2` (python binding for libgit2 C library); no `subprocess` calls to git CLI and no more massive string parsing in python
   - local git diff is retrieved and parsed by C library
   - diff returned from github REST API is parse by C library
- Performant lookup of file info (using `FileObj` class instead of a generic `dict`)
   - line numbers of added lines are consolidated to a list of `range()` parameters only once (before this was done multiple times for each file)
   - also simplifies typing of file info


2bndy5:
I'm not sure how to cover `pygit2` usage in CI since `actions/checkout` uses a shallow checkout by default. I'll have to think on this for a day...

Looking at the unit test runtime [before](https://github.com/cpp-linter/cpp-linter/actions/runs/7189817430/usage) vs [after](https://github.com/cpp-linter/cpp-linter/actions/runs/7190429900/usage), it looks like the new `FileObj` approach runs a few seconds faster in all scenarios.

cpp_linter/run.py
```diff
@@ -214,17 +207,28 @@ def verify_files_are_present() -> None:
...
+            raise ValueError(
+                "Could not clone repo into working directory because it is not empty"
+            ) from exc
+        commit = repo.revparse_single(GITHUB_SHA)
+        repo.checkout_tree(
+            cast(pygit2.Commit, commit).tree,
+            # reset index to specified commit
+            strategy=pygit2.GIT_CHECKOUT_FORCE | pygit2.GIT_CHECKOUT_RECREATE_MISSING,
+        )
+        repo.set_head(commit.oid)
```
2bndy5, line 231: This is the only change I'm concerned with. If the user does not checkout the repo but puts files into the `github.workspace` folder before running cpp-linter, then this change will crash the code.

I like this idea because
1. It makes us less dependent on the Github REST API
2. Instead of downloading only changed files, we get the entire snapshot of the repo. This should make clang-tidy happier when unchanged files are still present.
3. It encourages the user to checkout the repo beforehand if the code crashes here.

If this change is not reverted, then I'd like to figure out how to clone a shallow copy at the specified commit. Currently, this doesn't clone a shallow copy.

2bndy5: This code failed to communicate to server in CI, thus you had to retry the failed job. Due to this fact, I'm cherry-picking this change to another branch. Maybe we can revisit this idea when I explore gitlab support.

PR #41, opened 2023-12-14
fix annotations error counter

HerrCai0907:


2bndy5:
I noticed this today too. These changes are also part of #40. Hopefully I can rebase onto this.

2bndy5 (approved):

PR #47, opened 2024-01-01
make verbosity disabled by default

2bndy5:
Users can still enable verbosity by [re-running a CI job with debugging logs enabled](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging).

Not outputting debug info in production will save users CI time.

### DIFF
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -1,6 +1,7 @@
 """Setup the options for CLI arguments."""
 import argparse
 import logging
+from os import environ
 
 cli_arg_parser = argparse.ArgumentParser(
     description="Run clang-tidy and clang-format on a list of changed files "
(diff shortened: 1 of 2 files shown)


2bndy5:
speed improves by a negligible amount (~20 ms):

image

---
I'll be setting up a benchmark test in the cpp-linter/cpp_linter_rs repo that will use more complex and laborious scenarios. Currently, I'm just running the benchmark tests in repo root (using tests/demo/demo.* files and default checks).

2bndy5:
I have another speed-up trick up my sleeve: Parse clang-tidy and clang-format stdout from buffered strings instead of writing/reading to a cached files.

I only used cached files for easier development debugging. I rarely use the cached files for anything now. Unfortunately, this isn't a simple change given how the markdown comment is constructed from the parsed output.

PR #71, opened 2024-02-12
bump codecov-action to v4

2bndy5:
uses node.js v20 instead of deprecated node.js v16

also uses token in an env var as recommended by codecov docs


2bndy5:
I thought it was enabled for this repo, but I guess it is not. No wonder I had to bump these actions myself 🤣

2bndy5:
- [x] added the depndabot config.

Issue #75, opened 2024-02-13
PR review may not account for problems with no suggested fix


2bndy5:
Looking at the [list of supported clang-tidy diagnostics](https://clang.llvm.org/extra/clang-tidy/checks/list.html), there are many diagnostics that cannot be fixed automatically, only those with "Yes" in the "Offers fixes" column of the table.

This could be compensated by adding to the parsing of clang-tidy output.

<details><summary>example output</summary>

```
clang-tidy -p build -checks='cppcoreguidelines-*,clang-analyzer-*,modernize-*' --fix-errors src/demo.cpp src/demo.hpp
4025 warnings generated.
4030 warnings generated.
C:\dev\test-cpp-linter-action\src\demo.cpp:3:10: warning: inclusion of deprecated C++ header 'stdio.h'; consider using 'cstdio' instead [modernize-deprecated-headers]
#include <stdio.h>
         ^~~~~~~~~
         <cstdio>
C:\dev\test-cpp-linter-action\src\demo.cpp:3:10: note: FIX-IT applied suggested code changes
C:\dev\test-cpp-linter-action\src\demo.cpp:8:5: warning: use a trailing return type for this function [modernize-use-trailing-return-type]
int main(){
~~~ ^
auto       -> int
C:\dev\test-cpp-linter-action\src\demo.cpp:8:1: note: FIX-IT applied suggested code changes
int main(){
^
C:\dev\test-cpp-linter-action\src\demo.cpp:8:11: note: FIX-IT applied suggested code changes
int main(){
          ^
C:\dev\test-cpp-linter-action\src\demo.cpp:10:13: warning: statement should be inside braces [readability-braces-around-statements]
    for (;;) break;
            ^
             {
C:\dev\test-cpp-linter-action\src\demo.cpp:10:13: note: FIX-IT applied suggested code changes
C:\dev\test-cpp-linter-action\src\demo.cpp:10:20: note: FIX-IT applied suggested code changes
    for (;;) break;
                   ^
C:\dev\test-cpp-linter-action\src\demo.cpp:13:5: warning: do not call c-style vararg functions [cppcoreguidelines-pro-type-vararg]
    printf("Hello world!\n");
    ^
C:\dev\test-cpp-linter-action\src\demo.hpp:6:11: warning: use default member initializer for 'useless' [modernize-use-default-member-init]
    char* useless;
          ^
                 {"\0"}
[...]

PR #76, opened 2024-02-13
account for clang-tidy notes without fixes in PR reviews

2bndy5:
See #75 for detailed analysis. Basically, this

- runs clang-tidy once (instead of twice when `--tidy-review` is enabled)
- parses both notification and applied fixes from clang-tidy's stdout
- ~tries to correct the listed diagnostics about suggestions posted in PR review comments~ (see https://github.com/cpp-linter/cpp-linter/pull/76#issuecomment-1942205988)
- adds a separate PR review comment for each `TidyNotification` that had no fix applied.


2bndy5:
> tries to correct the listed diagnostics about suggestions posted in PR review comments

This didn't work as I hoped.

<details><summary>sample clang-tidy stdout</summary>

```
C:\dev\test-cpp-linter-action\src\demo.cpp:8:5: warning: use a trailing return type for this function [modernize-use-trailing-return-type]
int main(){
~~~ ^
auto       -> int
C:\dev\test-cpp-linter-action\src\demo.cpp:8:1: note: FIX-IT applied suggested code changes
int main(){
^
C:\dev\test-cpp-linter-action\src\demo.cpp:8:11: note: FIX-IT applied suggested code changes
int main(){
          ^
``` 

</details> 

So, Apparently the "FIX_IT applied ..." notes use the line numbers that correspond to the unmodified source code. I had hoped that the line number corresponded to the new modified source.

I'm still ok with this, but some of the listed diagnostic may not always get paired with the applicable suggestion.

2bndy5:
> adds a separate PR review comment for each TidyNotification that had no fix applied.

This worked!
image
But I'm going to add line and column info to the comment.

PR #79, opened 2024-02-21
Switch to reusable workflows

shenxianpeng:
As per https://github.com/cpp-linter/.github/issues/14

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -17,3 +17,4 @@ on:
 jobs:
   deploy:
     uses: cpp-linter/.github/.github/workflows/py-publish.yml@main
+    secrets: inherit


.github/workflows/build-docs.yml
```diff
@@ -3,31 +3,7 @@ name: Docs
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - uses: actions/setup-python@v5
-        with:
-          python-version: 3.x
-
-      - name: Install docs dependencies
-        run: pip install -r docs/requirements.txt -e .
```
2bndy5, line 16: The docs here require the cpp-linter package installed because the CLI doc is generated from cpp-linter/cli.py module. Thus the `-e .` here.

2bndy5:
I brute forced it instead:
https://github.com/cpp-linter/.github/blob/6cfb882552cd80e1f5f7d31e9726d94de60ed18f/.github/workflows/sphinx.yml#L27-L39

2bndy5:
BTW, to run changes in a reusable workflow, just select "re-run all jobs". Using "re-run job" will not use any updates in the reusable workflow.

2bndy5:
what about coverage reports?

.github/workflows/run-dev-tests.yml
```diff
@@ -104,47 +105,13 @@ jobs:
...
-          fail_ci_if_error: true # optional (default = false)
-          verbose: true # optional (default = false)
-
-      - name: Run codacy-coverage-reporter
-        uses: codacy/codacy-coverage-reporter-action@v1
-        with:
-          project-token: ${{ secrets.CODACY_PROJECT_TOKEN }}
-          coverage-reports: ./coverage.xml
+    uses: cpp-linter/.github/.github/workflows/py-coverage.yml@main
+  # Can not get coverage.xml from above reusable workflows.
```
2bndy5, line 109: I don't really care about uploading coverage reports to codacy; it was just an experiment of  mine. But If we wanted to, then the reusable workflow _could_ upload the coverage.xml as an artifact that we could use after the reusable workflow is finished.

2bndy5 (approved):
nice work!

PR #80, opened 2024-02-24
handle REST API rate limits and pagination

2bndy5:
- Implemented REST API rate limit avoidance as advised by [github REST API docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api).
- Implemented REST API pagination as advised by [github REST API docs](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api).

## pre-commit changes
- removed unnecessary dependencies needed to run mypy
- switch mypy hook to official mypy hook
- switch from using black to ruff for formatting

## Other changes
- make any critical HTTP requests that fail raise a `HTTPError` and removed some corresponding log output
- updated the module doc string in setup.py
- use Github REST API endpoint to download missing files (and keep their original encoding). Previously we were using raw URLs to download files, but that wasn't good for private repos because the `GITHUB_TOKEN` provided only applies to REST API calls.
- use `GITHUB_TOKEN` for all REST API calls (if provided)

### DIFF
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -338,7 +338,7 @@ class GithubApiClient(RestApiClient):
 
         :returns: If updating a comment, this will return the comment URL.
         """
-        logger.info("comments_url: %s", comments_url)
+        logger.debug("comments_url: %s", comments_url)
         comment_url: Optional[str] = None
         page = 1
         next_page: Optional[str] = comments_url + f"?page={page}&per_page=100"
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -19,7 +19,6 @@ from typing import Dict, List, Any, cast, Optional, Tuple, Union, Sequence
 
 from pygit2 import Patch  # type: ignore
 import requests
-from requests.structures import CaseInsensitiveDict
 from ..common_fs import FileObj, CACHE_PATH
 from ..clang_tools.clang_format import FormatAdvice, formalize_style_name
 from ..clang_tools.clang_tidy import TidyAdvice
@@ -57,6 +56,8 @@ class GithubApiClient(RestApiClient):
         self._rate_limit_remaining = -1  # -1 means unknown
         # a counter for avoiding secondary rate limits
         self._rate_limit_back_step = 0
+        # the rate limit reset time
+        self._rate_limit_reset: Optional[time.struct_time] = None
 
     def set_exit_code(
         self,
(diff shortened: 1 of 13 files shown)


cpp_linter/rest_api/github_api.py
```diff
@@ -66,6 +76,66 @@ def set_exit_code(
...
+    ):
+        logger.error("RATE LIMIT EXCEEDED!")
+        if "x-ratelimit-reset" in response_headers:
+            reset_epoch = time.gmtime(
+                int(cast(str, response_headers.get("x-ratelimit-reset")))
+            )
+            logger.error(
+                "Github REST API rate limit resets on %s",
+                time.strftime("%d %B %Y %H:%M +0000", reset_epoch),
+            )
```
2bndy5, line 93: I should probably exit early with `sys.exit(1)` to highlight the rate limit failure. Currently, the action will still pass with these logged errors, but that would produce a LGTM result when the rate limit is exceeded.

2bndy5 (commented):
I have no way of testing this in the test repo. It would take too much time and effort to hit the rate limit (5000 requests per hour using GITHUB_TOKEN) on purpose.

cpp_linter/rest_api/github_api.py
2bndy5: done

2bndy5:
This change also starts the foundation for [properly using REST API pagination](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api). Currently the API requests that _could_ return multiple page responses are for dismissing PR reviews. Deleting/updating thread comments already uses a countdown from total thread comments, but it costs us an extra API request just to get that total count of thread comments.

2bndy5:
I added proper pagination handling as well. These changes should help keep us from hitting REST API rate limits while supporting the full potential offered by the REST API's response headers.

cpp_linter/rest_api/github_api.py
```diff
@@ -67,6 +78,58 @@ def set_exit_code(
...
+        sys.exit(1)
+
+    def api_request(
+        self,
+        url: str,
+        method: Optional[str] = None,
+        data: Optional[str] = None,
+        headers: Optional[Dict[str, Any]] = None,
+    ) -> requests.Response:
+        if self._rate_limit_back_step > 5 or self._rate_limit_remaining == 0:
```
2bndy5, line 101:
```suggestion
        if self._rate_limit_back_step >= 5 or self._rate_limit_remaining == 0:
```

cpp_linter/rest_api/github_api.py
```diff
@@ -67,6 +78,58 @@ def set_exit_code(
...
+                logger.warning(
+                    "SECONDARY RATE LIMIT HIT! Backing off for %f seconds",
+                    wait_time,
+                )
+                time.sleep(wait_time)
+                self._rate_limit_back_step += 1
+                return self.api_request(url, method=method, data=data, headers=headers)
+            # primary rate limit handling
+            if self._rate_limit_remaining == 0:
+                self._rate_limit_exceeded(response.headers)
```
2bndy5, line 129: This code is executed immediately after a request response has been made. the repsonse's body could still be usable even if there are no allowed requests remaining.

Remove this check in favor of the same check located at the beginning of this function. In doing so, we can combine the `_rate_limit_exceeded()` code into this function because it would only get called from one location.

2bndy5:
# What if rate limits prevent us from getting the diff?
This thought spawned a deep dive into how we're using HTTP responses. In the end, I chose to use [`Response.raise_for_status()`](https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status) to propagate failed HTTP requests to the runner's log output. Using `raise_for_status()` helps us

1. highlight permission problems if user workflow has not granted us the permission to do what we need.
2. avoid logging erroneous requests.
3. ensure the responses' contents are used as expected

2bndy5:
Using [`Response.raise_for_status()`](https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status) could be a perceivable breaking change. Currently, some requests are allowed to fail silently.

I pushed a release to get #81 into cpp-linter-action (overhauling the mkdocs files right now). When I get around to thoroughly testing this, I'll do another release.

2bndy5:
I made the `raise_for_status()` call conditional using a `strict` arg to `api_request()`. This should be less obstructive and preserve most of original behavior.

> [!CAUTION]
> - `raise_for_status()` is still called for any HTTP request that is critical to expected behavior. ie. getting file changes, downloading missing files, posting/updating comments.
> - `raise_for_status()` is not called when deleting comments, dismissing reviews, or posting reviews.
>
> Posting reviews could be added to the list of critical operation for expected behavior in the future, but it still needs to be fleshed out.

2bndy5:
Both unit tests and integration tests (in our test repo) inspire confidence in that no expected behavior is broken. I'm ready to merge this on move on.

PR #49, opened 2024-01-02
complete refactor

2bndy5:
This uses a much more object-oriented approach.

- Functions are properly organized in modules with appropriate names.
- REST API usage is isolated in class derived from an abstract base class that will easily allow integrating other git servers' REST API.
- clang-format output is filtered by lines (when applicable) as the XML is parsed.
- The comment is made after all clang tool output is parsed.
- All clang tool output is parsed from buffered strings (addresses #48).
- There isn't any cached files when verbosity is not set to debug mode.
- Tests can now use mocked HTTP requests (no more fear of hitting REST API rate limits).
- Updated docs to reflect the new package structure.
- Removed dead code (from previous attempts to generate PR reviews).
- Added ability to pass positional/optional CLI arguments. These arguments will be treated as paths and automatically included in the `not_ignored` list of files. However, the paths are still subject to further filtering with `--extensions`, `--lines-changed-only`, and `--files-changed-only` options.

PR #81, opened 2024-02-24
re-enable thread-comments on private repos

2bndy5:
I took another swing at this. It turns out the media type used in [request headers needs to select `raw+json`](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#list-issue-comments-for-a-repository). This way we can see the raw contents of comments for a PR.

I also disabled saving the list of comments to a json file when logging is not verbose. This should've been changed in #47 or in #49, but I must have missed it.


2bndy5:
I have actually tested this in a private repo of mine. I'm also using that private repo to test a similar contribution to a different [arduino/report-size-deltas](https://github.com/arduino/report-size-deltas) action that posts comments to a PR.

PR #85, opened 2024-03-01
Revise CLI doc generation

2bndy5:
adds permission links and required version for each input option

### DIFF
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -56,7 +56,7 @@ cli_arg_parser.add_argument(
     "-s",
     "--style",
     default="llvm",
-    help="""The style rules to use (defaults to ``%(default)s``).
+    help="""The style rules to use.
 
 - Set this to ``file`` to have clang-format use the
   closest relative .clang-format file.
@@ -64,7 +64,8 @@ cli_arg_parser.add_argument(
   using clang-format entirely.
 
 See `clang-format docs <https://clang.llvm.org/docs/ClangFormat.html>`_ for more info.
-""",
+
+Defaults to ``%(default)s``""",
 )
 cli_arg_parser.add_argument(
     "-c",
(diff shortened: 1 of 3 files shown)


2bndy5:
You can now see the doc builds previewed at RTD.
image
Just click "Details".

2bndy5:
The navigation bar is the same position 😕 for each. The page break points might be different, but that just depends on how wide your browser window is.

Or are you talking about the RTD menu? 
image

2bndy5:
Aside from background color in dark mode, I don't see a difference. The nav bar should appear on top of page. The nav menu (technically called a "drawer" on mobile devices) should appear from the left. On desktop with browser window as wide as possible, the **page**'s Table of Contents (ToC) would appear on the right, but the **site**'s ToC would appear in the nav menu on the left. Both **page** and **site** ToC would appear combined on mobile devices.

It may take pictures to explain what you're trying to say.

The sphinx-immaterial theme is not up to date with changes from it's upstream mkdocs-material theme. We have been slacking on that because mkdocs-material's theme source files suffered a significant refactor (which wouldn't affect users' experience); something like 300 file changes would have to get diligently reviewed in a single PR to sphinx-immaterial.

2bndy5:
Oh, you're talking about the [navigation tabs](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-tabs). That is a feature that we'd have enable in [cpp-linter-action/mkdocs.yml](https://github.com/cpp-linter/cpp-linter-action/blob/d4202dcac7e45400beb7d6b6359929734ef2b3e6/mkdocs.yml#L17).

2bndy5:
As for the left side ToC in cpp-linter, I'll have to look into that. I think it is a Sphinx problem; the way the ToC is generated by sphinx is very different from mkdocs.

2bndy5:
> As for the left side ToC in cpp-linter, I'll have to look into that. I think it is a Sphinx problem; the way the ToC is generated by sphinx is very different from mkdocs.

I can't make them exactly behave the same because sphinx is different from mkdocs. Mkdocs puts everything nested under the root page. Sphinx does not nest everything under the root page.

2bndy5:
I had to disable navigation tabs to get the effect you desire. I think this might have been an opinionated implementation by the sphinx-immaterial author (jbms, not me). Or maybe it just a difference implemented in mkdocs-material after v8.5.6.

2bndy5:
I'm ready to merge this. The unit test for CLI args passes under the same expectations as before. So, there are no actual API changes here.

PR #87, opened 2024-03-08
fix autolabeler by adding labeler.yml action

shenxianpeng:
Per as https://github.com/cpp-linter/.github/issues/21

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -4,7 +4,6 @@ on:
   # pull_request event is required for autolabeler
   pull_request:
     types: [opened, reopened, synchronize]
-  workflow_dispatch:
 
 jobs:
   draft-release:


2bndy5:
Currently assessing doc build failures in jbms/sphinx-immaterial#333

We might have to use a cache of fonts for now... From a different PR.

.github/workflows/release-drafter.yml
```diff
@@ -4,8 +4,17 @@ on:
   push:
     branches:
       - "main"
+  # pull_request event is required only for autolabeler
+  pull_request:
+    # Only following types are handled by the action, but one can default to all as well
+    types: [opened, reopened, synchronize]
+  # pull_request_target event is required for autolabeler to support PRs from forks
+  pull_request_target:
+    types: [opened, reopened, synchronize]
```
2bndy5, line 13: Is there a reason you want to trigger on both `pull_request` and `pull_request_target` events?

It seems that `pull_request_target` would suite us just fine for this action.

2bndy5: It also works for origin remotes. I think this would lead to duplicate CI runs if `pull_request` trigger is also enabled.

The only real difference (for security reasons) is that `pull_request_target` runs in the context of the PR's target branch.

As I understand it, release-drafter should only need to be run on commits to main branch. In fact, there have been a couple times where I forgot to add the right labels to a PR. So, I added the labels to the closed PR and re-ran release-drafter (that was triggered by push to main branch), and the drafted release was corrected.

2bndy5:
TL;DR We need to use a different action to generate labels in a PR.

The [issue you linked to](https://github.com/release-drafter/release-drafter/issues/1125) has diverged into 2 issues.
1. release-drafter cannot create a drafted release in a PR event (including `pull_request_target`). This is why the error occurs.
2. release-drafter's auto-labeler feature needs to be run in a PR event.

But there is currently no way differentiate which feature of release-drafter you want to use in specified event trigger. Using the `commitish` advice will fix the error in a PR event, but it will prevent auto-labeler from doing what it supposed to.

.github/workflows/release-drafter.yml
2bndy5: > It also works for origin remotes. I think this would lead to duplicate CI runs 

I was wrong. `pull_request_target` didn't get triggered on this branch because it shares the same remote.

See also [my full assessment](https://github.com/cpp-linter/cpp-linter/pull/87#issuecomment-1984978713)

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,11 @@
+name: Autolabeler
```
2bndy5, line 1:
```suggestion
name: PR Autolabeler
```
For better distinction in the list of all workflows

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on:
+  # pull_request event is required for autolabeler
+  pull_request:
+    types: [opened, reopened, synchronize]
+  workflow_dispatch:
```
2bndy5, line 7: The `workflow_dispatch` won't be useful here because the auto-labeler needs to be used in a `pull_request` event. Please remove this trigger.

2bndy5 (changes requested):
I'm surprised that this works without errors. It seems to use the `commitish` advice (in reusable workflow), but the labels (ie "bug" in this PR) are still assigned to the PR properly.

.github/workflows/labeler.yml
2bndy5:
```suggestion
```

2bndy5 (approved):

2bndy5:
Is that why `commitish` is hard-coded in the reusable workflow?

PR #92, opened 2024-03-20
Enable parallelism

### LINKED ISSUE #74: Parallelize lint checks
The lint workflow only processes a single file at a time. For a large project in which most PRs touch multiple files this can result in the workflow taking up to a hour or longer. People allocating more vcpus to their runner or running self-hosted runners using server grade hardware (e.g. 2+ CPUs with 20+ cores per CPU) can benefit from running multiple lint jobs at the same time similar to passing a -j parameter to clang-tidy.

jnooree:
Closes #74. I also reviewed #73.

### DIFF
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -70,7 +70,7 @@ def _run_on_single_file(
             format_cmd, file, style, lines_changed_only, format_review
         )
 
-    return log_file, tidy_note, format_advice
+    return file.name, log_file, tidy_note, format_advice
 
 
 def capture_clang_tools_output(
@@ -131,11 +131,8 @@ def capture_clang_tools_output(
         if db_path.exists():
             db_json = json.loads(db_path.read_text(encoding="utf-8"))
 
-    # temporary cache of parsed notifications for use in log commands
-    tidy_notes = []
-    format_advice = []
-    log_lvl = logger.getEffectiveLevel()
     with TemporaryDirectory() as temp_dir, ProcessPoolExecutor(num_workers) as executor:
+        log_lvl = logger.getEffectiveLevel()
         futures = [
             executor.submit(
                 _run_on_single_file,
@@ -156,16 +153,20 @@ def capture_clang_tools_output(
             for file in files
         ]
 
-        for file, future in zip(files, as_completed(futures)):
-            log_file, note, advice = future.result()
+        # temporary cache of parsed notifications for use in log commands
+        format_advice_map: Dict[str, Optional[FormatAdvice]] = {}
+        tidy_notes_map: Dict[str, Optional[TidyAdvice]] = {}
+        for future in as_completed(futures):
+            file, log_file, note, advice = future.result()
 
-            start_log_group(f"Performing checkup on {file.name}")
+            start_log_group(f"Performing checkup on {file}")
             sys.stdout.write(Path(log_file).read_text())
             end_log_group()
 
-            if note is not None:
-                tidy_notes.append(note)
-            if advice is not None:
-                format_advice.append(advice)
+            format_advice_map[file] = advice
+            tidy_notes_map[file] = note
+
+    format_advice = list(filter(None, (format_advice_map[file.name] for file in files)))
+    tidy_notes = list(filter(None, (tidy_notes_map[file.name] for file in files)))
 
     return (format_advice, tidy_notes)
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -51,8 +51,6 @@ def run_on_single_file(
 ):
     logfile = worker_logfile(tempdir)
 
-    start_log_group(f"Performing checkup on {file.name}")
-
     tidy_note = None
     if tidy_cmd is not None:
         tidy_note = run_clang_tidy(
@@ -71,8 +69,6 @@ def run_on_single_file(
         format_advice = run_clang_format(
             format_cmd, file, style, lines_changed_only, format_review
         )
-
-    end_log_group()
 
     return logfile, tidy_note, format_advice
 
@@ -155,8 +151,11 @@ def capture_clang_tools_output(
             files,
         )
 
-        for logfile, note, advice in results:
+        for file, (logfile, note, advice) in zip(files, results):
+            start_log_group(f"Performing checkup on {file.name}")
             sys.stdout.write(Path(logfile).read_text())
+            end_log_group()
+
             if note is not None:
                 tidy_notes.append(note)
             if advice is not None:
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -1,12 +1,16 @@
+from functools import partial
 import json
+from multiprocessing import Pool
 from pathlib import Path, PurePath
 import subprocess
+import sys
+from tempfile import TemporaryDirectory
 from textwrap import indent
 from typing import Optional, List, Dict, Tuple
 import shutil
 
 from ..common_fs import FileObj
-from ..loggers import start_log_group, end_log_group, logger
+from ..loggers import start_log_group, end_log_group, worker_logfile, logger
 from .clang_tidy import run_clang_tidy, TidyAdvice
 from .clang_format import run_clang_format, FormatAdvice
 
@@ -31,6 +35,48 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
     return shutil.which(tool_name)
 
 
+def run_on_single_file(
+    file: FileObj,
+    tempdir: str,
+    tidy_cmd,
+    checks,
+    lines_changed_only,
+    database,
+    extra_args,
+    db_json,
+    tidy_review,
+    format_cmd,
+    style,
+    format_review,
+):
+    logfile = worker_logfile(tempdir)
+
+    start_log_group(f"Performing checkup on {file.name}")
+
+    tidy_note = None
+    if tidy_cmd is not None:
+        tidy_note = run_clang_tidy(
+            tidy_cmd,
+            file,
+            checks,
+            lines_changed_only,
+            database,
+            extra_args,
+            db_json,
+            tidy_review,
+        )
+
+    format_advice = None
+    if format_cmd is not None:
+        format_advice = run_clang_format(
+            format_cmd, file, style, lines_changed_only, format_review
+        )
+
+    end_log_group()
+
+    return logfile, tidy_note, format_advice
+
+
 def capture_clang_tools_output(
     files: List[FileObj],
     version: str,
@@ -41,6 +87,7 @@ def capture_clang_tools_output(
     extra_args: List[str],
     tidy_review: bool,
     format_review: bool,
+    num_workers: int,
 ) -> Tuple[List[FormatAdvice], List[TidyAdvice]]:
     """Execute and capture all output from clang-tidy and clang-format. This aggregates
     results in the :attr:`~cpp_linter.Globals.OUTPUT`.
@@ -89,26 +136,30 @@ def capture_clang_tools_output(
     # temporary cache of parsed notifications for use in log commands
     tidy_notes = []
     format_advice = []
-    for file in files:
-        start_log_group(f"Performing checkup on {file.name}")
-        if tidy_cmd is not None:
-            tidy_notes.append(
-                run_clang_tidy(
-                    tidy_cmd,
-                    file,
-                    checks,
-                    lines_changed_only,
-                    database,
-                    extra_args,
-                    db_json,
-                    tidy_review,
-                )
-            )
-        if format_cmd is not None:
-            format_advice.append(
-                run_clang_format(
-                    format_cmd, file, style, lines_changed_only, format_review
-                )
-            )
-        end_log_group()
+    with TemporaryDirectory() as tempdir, Pool(num_workers) as pool:
+        results = pool.imap(
+            partial(
+                run_on_single_file,
+                tempdir=tempdir,
+                tidy_cmd=tidy_cmd,
+                checks=checks,
+                lines_changed_only=lines_changed_only,
+                database=database,
+                extra_args=extra_args,
+                db_json=db_json,
+                tidy_review=tidy_review,
+                format_cmd=format_cmd,
+                style=style,
+                format_review=format_review,
+            ),
+            files,
+        )
+
+        for logfile, note, advice in results:
+            sys.stdout.write(Path(logfile).read_text())
+            if note is not None:
+                tidy_notes.append(note)
+            if advice is not None:
+                format_advice.append(advice)
+
     return (format_advice, tidy_notes)
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -10,7 +10,7 @@ from typing import Optional, List, Dict, Tuple
 import shutil
 
 from ..common_fs import FileObj
-from ..loggers import start_log_group, end_log_group, worker_logfile, logger
+from ..loggers import start_log_group, end_log_group, worker_logfile_init, logger
 from .clang_tidy import run_clang_tidy, TidyAdvice
 from .clang_format import run_clang_format, FormatAdvice
 
@@ -38,6 +38,7 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
 def run_on_single_file(
     file: FileObj,
     tempdir: str,
+    loglvl: int,
     tidy_cmd,
     checks,
     lines_changed_only,
@@ -49,7 +50,7 @@ def run_on_single_file(
     style,
     format_review,
 ):
-    logfile = worker_logfile(tempdir)
+    logfile = worker_logfile_init(tempdir, loglvl)
 
     tidy_note = None
     if tidy_cmd is not None:
@@ -132,11 +133,13 @@ def capture_clang_tools_output(
     # temporary cache of parsed notifications for use in log commands
     tidy_notes = []
     format_advice = []
+    loglvl = logger.getEffectiveLevel()
     with TemporaryDirectory() as tempdir, Pool(num_workers) as pool:
         results = pool.imap(
             partial(
                 run_on_single_file,
                 tempdir=tempdir,
+                loglvl=loglvl,
                 tidy_cmd=tidy_cmd,
                 checks=checks,
                 lines_changed_only=lines_changed_only,
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -35,7 +35,7 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
     return shutil.which(tool_name)
 
 
-def run_on_single_file(
+def _run_on_single_file(
     file: FileObj,
     tempdir: str,
     loglvl: int,
@@ -137,7 +137,7 @@ def capture_clang_tools_output(
     with TemporaryDirectory() as tempdir, Pool(num_workers) as pool:
         results = pool.imap(
             partial(
-                run_on_single_file,
+                _run_on_single_file,
                 tempdir=tempdir,
                 loglvl=loglvl,
                 tidy_cmd=tidy_cmd,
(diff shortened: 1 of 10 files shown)


cpp_linter/loggers.py
```diff
@@ -53,3 +54,29 @@ def log_response_msg(response: Response):
             response.request.url,
             response.text,
         )
+
+
+def worker_logfile(tempdir: str):
+    logfile = NamedTemporaryFile("w", dir=tempdir, delete=False)
```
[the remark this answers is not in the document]
jnooree: Yes, I implemented with StringIO first then fall back to the current version. For large repositories with many logs might explode memory.

This version is still *much* faster than the serial version (also pointed by #37). The clang-tidy check is slow enough to use the disk.

tests/capture_tools_output/test_database_path.py
```diff
@@ -53,23 +54,19 @@ def test_db_detection(
         extra_args=[],
         tidy_review=False,
         format_review=False,
+        num_workers=2,
```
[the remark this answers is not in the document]
jnooree: Correct, I just wanted to ensure multiple processes to be spawned. One process will run clang-tidy and the other will just sleep, so might not be an issue?

tests/capture_tools_output/test_tools_output.py
```diff
@@ -295,8 +298,6 @@ def test_format_annotations(
                         break
                 else:  # pragma: no cover
                     raise RuntimeError(f"line {line} not in ranges {repr(ranges)}")
-        else:  # pragma: no cover
-            raise RuntimeWarning(f"unrecognized record: {message}")
```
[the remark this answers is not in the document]
jnooree: Now that the loop is unrolled, not required anymore.

tests/capture_tools_output/test_database_path.py
```diff
@@ -53,23 +54,19 @@ def test_db_detection(
...
-    for record in caplog.records:
-        assert "Error while trying to load a compilation database" not in record.message
-        msg_match = CLANG_TIDY_COMMAND.search(record.message)
-        if msg_match is not None:
-            matched_args = msg_match.group(0).split()[1:]
-            break
-    else:  # pragma: no cover
+    stdout = capsys.readouterr().out
+    assert "Error while trying to load a compilation database" not in stdout
+    msg_match = CLANG_TIDY_COMMAND.search(stdout)
```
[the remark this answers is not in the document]
jnooree: This is the hardest part to solve in this attempt. I also checked with rich installed, but the file location was not included in the parsed arguments for my terminal size...

How about temporarily disabling rich in tests? This test wants to verify that the arguments are passed correctly, not the log output format is correct, anyways.

cpp_linter/loggers.py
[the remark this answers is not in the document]
2bndy5: > For large repositories with many logs might explode memory.

I can't argue with that (scalability). Maybe in the future, I'll use temp files as a fallback when the list of total line changes is over 100. I would still like to use in-memory streams for small repos, but I guess the difference would be negligible. And there's no telling how many clang-tidy diagnostics could be triggered for each line.

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
2bndy5: The test should reflect the performance of the machine running it. And since we're not actually testing `main()`, then I think it is appropriate to set this dynamically, even if only 1 process is used.

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
2bndy5: > How about temporarily disabling rich in tests?

I'm ok with that. The hard part is the fact that `capture_clang_tools_output()` indirectly calls `worker_logfile_init()` which configures the logger depending on rich pkg's presence. How would you go about disabling rich for tests? I guess we could use a env var that the test `monkeypatch.setenv()`.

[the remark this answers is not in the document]
2bndy5:
```python
        handler = RichHandler(
            show_time=False,
            console=console,
            show_path="CPP_LINTER_PYTEST_RUN" not in os.environ
        )
```

2bndy5:
# 🎉 Benchmarking at repo root on my intel i7-9700 CPU
```
hyperfine "cpp-linter -l false -a false -j 1" "cpp-linter -l false -a false -j 0"
Benchmark 1: cpp-linter -l false -a false -j 1
  Time (mean ± σ):     51.437 s ±  2.264 s    [User: 25.653 s, System: 12.871 s]
  Range (min … max):   49.109 s … 54.378 s    10 runs

Benchmark 2: cpp-linter -l false -a false -j 0
  Time (mean ± σ):     26.391 s ±  0.498 s    [User: 26.286 s, System: 12.340 s]
  Range (min … max):   25.500 s … 27.063 s    10 runs

Summary
  cpp-linter -l false -a false -j 0 ran
    1.95 ± 0.09 times faster than cpp-linter -l false -a false -j 1
```

cpp_linter/loggers.py
```diff
@@ -53,3 +54,32 @@ def log_response_msg(response: Response):
...
+    logger.addHandler(handler)
+    # Windows does not copy log level to subprocess.
+    # https://github.com/cpp-linter/cpp-linter/actions/runs/8355193931
+    logger.setLevel(loglvl)
+
+    log_commander.handlers.clear()
+    log_commander.propagate = False
+    console_handler = logging.StreamHandler(logfile)
+    console_handler.setFormatter(logging.Formatter("%(message)s"))
+    log_commander.addHandler(console_handler)
```
2bndy5, line 83: With the log groups done in the main thread, I don't think we need to re-configure `log_commander`, right?

The clang-tidy.py and clan-format.py modules don't actually use `log_commander` (directly or otherwise).

2bndy5 (commented):
I'm also noticing that you don't use a word delimiter `_` in your variable names (which should be snake-cased according to PEP8). This is throwing my spellchecker in a frenzy.

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
jnooree: Second version would be much simpler. Thanks for the suggestions!

cpp_linter/loggers.py
jnooree: Correct, I just wanted to make sure everything to be captured. Maybe at some point log_commander could be used while running tools?

jnooree:
> I'm also noticing that you don't use a word delimiter _ in your variable names (which should be snake-cased according to PEP8). This is throwing my spellchecker in a frenzy.

If you want to change some names, please locate the names so I could update them.

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
jnooree: Ok, then this should be done in the command line parser.

cpp_linter/loggers.py
2bndy5: > Maybe at some point log_commander could be used while running tools?

Its not likely. The log commander is only used to output github action log commands:
1. [grouping log lines](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines)
2. [making file annotations](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message)

I don't see any more usefulness [from those options](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions), especially during in threads when running clang-format and clang-tidy.

If you just want to comment the paragraph out, I'm ok with that too. Maybe github will add something that might be useful in the future. I just want to reduce the lines executed for each file's analysis.
```suggestion
    ## uncomment the following if log_commander is needed in isolated threads
    # log_commander.handlers.clear()
    # log_commander.propagate = False
    # console_handler = logging.StreamHandler(logfile)
    # console_handler.setFormatter(logging.Formatter("%(message)s"))
    # log_commander.addHandler(console_handler)
```

cpp_linter/clang_tools/__init__.py
```diff
@@ -31,6 +35,45 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
     return shutil.which(tool_name)
 
 
+def _run_on_single_file(
+    file: FileObj,
+    tempdir: str,
+    loglvl: int,
```
2bndy5, line 41:
```suggestion
    temp_dir: str,
    log_lvl: int,
```

cpp_linter/clang_tools/__init__.py
```diff
@@ -31,6 +35,45 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
...
+    lines_changed_only,
+    database,
+    extra_args,
+    db_json,
+    tidy_review,
+    format_cmd,
+    style,
+    format_review,
+):
+    logfile = worker_logfile_init(tempdir, loglvl)
```
2bndy5, line 53:
```suggestion
    log_file = worker_log_file_init(temp_dir, log_lvl)
```

cpp_linter/loggers.py
```diff
@@ -53,3 +54,32 @@ def log_response_msg(response: Response):
             response.request.url,
             response.text,
         )
+
+
+def worker_logfile_init(tempdir: str, loglvl: int):
+    logfile = NamedTemporaryFile("w", dir=tempdir, delete=False)
```
2bndy5, line 61:
```suggestion
def worker_log_file_init(temp_dir: str, log_lvl: int):
    log_file = NamedTemporaryFile("w", dir=temp_dir, delete=False)
```

cpp_linter/clang_tools/__init__.py
```diff
@@ -89,26 +133,35 @@ def show_tool_version_output(cmd: str):  # show version output for executable us
...
-                )
-            )
-        end_log_group()
+    loglvl = logger.getEffectiveLevel()
+    with TemporaryDirectory() as tempdir, Pool(num_workers) as pool:
+        results = pool.imap(
+            partial(
+                _run_on_single_file,
+                tempdir=tempdir,
+                loglvl=loglvl,
```
2bndy5, line 144:
```suggestion
                temp_dir=temp_dir,
                log_lvl=log_lvl,
```

cpp_linter/clang_tools/__init__.py
```diff
@@ -89,26 +133,35 @@ def show_tool_version_output(cmd: str):  # show version output for executable us
...
-            )
-        if format_cmd is not None:
-            format_advice.append(
-                run_clang_format(
-                    format_cmd, file, style, lines_changed_only, format_review
-                )
-            )
-        end_log_group()
+    loglvl = logger.getEffectiveLevel()
+    with TemporaryDirectory() as tempdir, Pool(num_workers) as pool:
```
2bndy5, line 139:
```suggestion
    with TemporaryDirectory() as temp_dir, Pool(num_workers) as pool:
```

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
jnooree: Applied in 45cfc6d360c4440ee83430c832f8616794462ec6.

cpp_linter/clang_tools/__init__.py
```diff
@@ -31,6 +35,45 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
...
+            tidy_review,
+        )
+
+    format_advice = None
+    if format_cmd is not None:
+        format_advice = run_clang_format(
+            format_cmd, file, style, lines_changed_only, format_review
+        )
+
+    return logfile, tidy_note, format_advice
```
2bndy5, line 74:
```suggestion
    return log_file, tidy_note, format_advice
```

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
jnooree: Fixed in 0e78226dcd10aa2b9df60e588a081d4cb15d434f.

tests/capture_tools_output/test_tools_output.py
```diff
@@ -274,12 +276,13 @@ def test_format_annotations(
...
     gh_client.make_annotations(files, format_advice, tidy_advice, style)
-    for message in [r.message for r in caplog.records if r.levelno == logging.INFO]:
-        if FORMAT_RECORD.search(message) is not None:
-            line_list = message[message.find("style guidelines. (lines ") + 25 : -1]
-            lines = [int(line.strip()) for line in line_list.split(",")]
+    logline: str
+    for logline in capsys.readouterr().err.splitlines():
+        if FORMAT_RECORD.search(logline) is not None:
+            line_list = logline[logline.find("style guidelines. (lines ") + 25 : -1]
+            lines = [int(logline.strip()) for logline in line_list.split(",")]
```
2bndy5, line 283:
```suggestion
    log_line: str
    for log_line in capsys.readouterr().err.splitlines():
        if FORMAT_RECORD.search(log_line) is not None:
            line_list = log_line[log_line.find("style guidelines. (lines ") + 25 : -1]
            lines = [int(log_line.strip()) for log_line in line_list.split(",")]
```

tests/capture_tools_output/test_tools_output.py
```diff
@@ -274,12 +276,13 @@ def test_format_annotations(
...
             file_obj = match_file_json(
-                RECORD_FILE.sub("\\1", message).replace("\\", "/"), files
+                RECORD_FILE.sub("\\1", logline).replace("\\", "/"), files
```
2bndy5, line 285:
```suggestion
                RECORD_FILE.sub("\\1", log_line).replace("\\", "/"), files
```

2bndy5 (changes requested):
It is odd to request a refactor from the reviewer. Maybe I didn't describe what I wanted (in accordance with PEP8).

I'm probably missing some spots. Renaming symbols is better done in an actual IDE, not via PR diff. 😆

tests/capture_tools_output/test_database_path.py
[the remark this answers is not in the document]
2bndy5: You picked a better env var name than the one I suggested 👍🏼

jnooree:
> It is odd to request a refactor from the reviewer. Maybe I didn't describe what I wanted (in accordance with PEP8).

Thanks for the suggestions! I don't use a spell checker, so I just wanted to know the spell checker output of your environment.

2bndy5:
I'm adding a pre-commit hook to run [codespell](https://github.com/codespell-project/codespell) in #93. I realized my request wasn't integrated into our CI checks. ruff is pretty relaxed about variable naming in a function's scope.

The actual spell checker that I'm using is a [VSCode extension](https://github.com/streetsidesoftware/vscode-spell-checker) (not the codespell package).

jnooree:
I just finished variable renaming! No more force pushes.

tests/capture_tools_output/test_tools_output.py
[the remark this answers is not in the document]
2bndy5: image
This coverage is exactly why that `else: raise` was there. the test is passing because there is nothing to parse. Something is wrong here. Maybe change to `capsys.readouterr().out` (not `.err`)?

cpp_linter/loggers.py
```diff
@@ -53,3 +55,33 @@ def log_response_msg(response: Response):
             response.request.url,
             response.text,
         )
+
+
+def worker_log_file_init(temp_dir: str, log_lvl: int):
```
2bndy5, line 60: for some reason this function is never executed in unit tests. See [coverage results](https://app.codecov.io/gh/cpp-linter/cpp-linter/pull/92/blob/cpp_linter/loggers.py)

2bndy5: I think the separate processes are messing with the code coverage analysis. The `_run_on_single_file()` function is [not showing coverage either](https://app.codecov.io/gh/cpp-linter/cpp-linter/pull/92/blob/cpp_linter/clang_tools/__init__.py).

2bndy5: Here is the [coverage docs about measuring subprocesses](https://coverage.readthedocs.io/en/7.4.4/subprocess.html). Not sure if that is helpful. There's also [this SO suggestion](https://stackoverflow.com/a/71023304), but we use the pyproject.toml to configure the coverage tool:
https://github.com/cpp-linter/cpp-linter/blob/005c30b0c7a989bde002bc3138ec5dcf131d6c22/pyproject.toml#L67-L76

2bndy5: Now I see that you pushed some changes to help measure the code coverage. Well done! 💯

tests/capture_tools_output/test_tools_output.py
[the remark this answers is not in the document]
2bndy5: disregard, coverage report was out of date.

tests/test_cli_args.py
```diff
@@ -77,13 +78,16 @@ def test_defaults():
         ("extra-arg", '"-std=c++17 -Wall"', "extra_arg", ['"-std=c++17 -Wall"']),
         ("tidy-review", "true", "tidy_review", True),
         ("format-review", "true", "format_review", True),
+        ("jobs", "0", "jobs", None),
+        ("jobs", "1", "jobs", 1),
+        ("jobs", "4", "jobs", 4),
```
2bndy5, line 83: To cover the exception raised in cli.py:
```suggestion
        ("jobs", "4", "jobs", 4),
        pytest.param("jobs", "x", "jobs", 0, marks=pytest.mark.xfail),
```

cpp_linter/cli.py
```diff
@@ -304,6 +305,31 @@
 )
 
 
+def _parse_jobs(val: str) -> Optional[int]:
+    try:
+        jobs = int(val)
+    except ValueError:  # pragma: no cover
+        raise argparse.ArgumentTypeError(f"invalid value: {val} (must be an integer)")
```
2bndy5, line 312: We can actually cover this in a unit test using `pytest.mark.xfail` (see other review comment). TBH, I think this whole function could be reduced to a lambda:
```diff
-    type=_parse_jobs,
+    type=lambda x: None if int(x) <= 0 else int(x),
```
But maybe its not good to rely on `int()` raising a generic looking `ValueError`.

Instead, we can raise `from` the caught `ValueError` to avoid a lengthy traceback output:
```suggestion
    except ValueError as exc:
        raise argparse.ArgumentTypeError(
            f"Invalid -j (--jobs) value: {val} (must be an integer)"
        ) from exc
```

2bndy5 (changes requested):
A few final coverage adjustments (to make me sleep better)

jnooree:
I think there are some suspicious missing lines in the coverage report. I will investigate them first.

jnooree:
The "missing" lines are actually covered in tests (checked manually), but coverage could not track the changes. Are we ok to skip them or need some workarounds?

2bndy5:
Well, I'm still puzzled why only the `lines_changed_only` arg isn't triggered in coverage.

I'd rather not skip the lines just to make coverage report "look" better. I prefer we actually have reproducible results to measure for code coverage.

jnooree:
I'll experiment for other methods to get coverage report correctly and share if any succeded. I think this looks like an upstream bug tho.

2bndy5:
I'm seeing a lot of .coverage files created in the tests folder. When I run `coverage combine`, these `tests/.coverage.*` files are not combined into the report. I think one of the tests is changing coverage's working directory.

2bndy5:
If I move all the rogue .coverage files into repo root and run `coverage combine`, then the output says it skips 146 "duplicate data"; only 6 .coverage files are actually combined.

This skipping "duplicate data" isn't a new thing, but spawning the coverage files in the tests folder is a new thing. I hesitate to call it a problem because I don't know the inner-workings of the coverage package.

jnooree:
Yes, I also just found out that the monkeypatch might cause problems for subprocesses. Fixed in 696d83c64b50acef5169a9bbb54eb882d7786b59.

2bndy5:
The rogue .coverage file are spawned when running
- tests/capture_tools_output/test_database_path.py

jnooree:
Oh, ok. It seems the tests are not very pleased with out-of-order logs. I'll update them shortly.

2bndy5:
I must say, you are picking up on the cause of these errors faster than I'm accustomed to. Its very refreshing!

2bndy5 (approved):
> Oh, ok. It seems the tests are not very pleased with out-of-order logs

That was it completely! coverage is restored and new functionality verified 💯

2bndy5:
@jnooree Thank you soooo much!❤️ I don't think I would've been able to solve #74 on my own.

jnooree:
I should thank you and the maintainers first for creating such a great package/workflow. Great work!

PR #93, opened 2024-03-20
add codespell to pre-commit hooks

2bndy5:
also updated pre-commit hooks and ran them

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,14 +15,14 @@ repos:
         args: ["--fix=lf"]
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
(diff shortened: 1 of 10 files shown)


2bndy5:
I'm going to merge #92 first, then rebase all my branches open in PRs...

@shenxianpeng BTW, this codespell pre-commit hook would be nice to use in cpp-linter-action and our other repos. Maybe the pre-commit reusable workflow could add [the codespell problem matcher](https://github.com/codespell-project/codespell-problem-matcher) as an extra setup step.

---
PS - I know there is a [github action](https://github.com/codespell-project/actions-codespell) that includes a problem matcher for the codespell project, but it has problems. Most notably, you cannot select what version of codespell it runs.

PR #94, opened 2024-03-23
perf: use io.StringIO instead tempdir/tempfile

jnooree:
This is a follow-up PR of #92.

In the previous PR review https://github.com/cpp-linter/cpp-linter/pull/92#discussion_r1532181842 @2bndy5 suggested:

> Can we avoid caching to a temp file? I raised some concern about this approach in [cpp-linter#73](https://github.com/orgs/cpp-linter/discussions/73#discussioncomment-8486098) as well (about cost in speed).
> Could we use [`io.StringIO` stream](https://docs.python.org/3/library/io.html#text-i-o) instead?

In that version, [multiprocessing.Pool.imap() was used](https://github.com/jnooree/cpp-linter/blob/5fefb726a6246cf00ab72f9045d9cea00d2e6163/cpp_linter/clang_tools/__init__.py#L139-L140) for parallel processing, and it [blocks](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.imap) until all of the results are made available (if I'm not misunderstood). The behavior might cause memory problems on repositories with many files and warnings.

After, I decided to switch to the async version in 4e029a317fe51a2399cd6519edd11422a8eb5448 by using [concurrent.futures.as_completed](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed), which returns the logs as each work is completed (hence its name). Now that the logs are immediately printed after the subprocess is done processing a file and logs for a single file would not overflow memory in most projects, I think it is safe to use `io.StringIO` for better performance (as with the original suggestion).

### DIFF
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -154,10 +154,10 @@ def capture_clang_tools_output(
         format_advice_map: Dict[str, Optional[FormatAdvice]] = {}
         tidy_notes_map: Dict[str, Optional[TidyAdvice]] = {}
         for future in as_completed(futures):
-            file, log, note, advice = future.result()
+            file, logs, note, advice = future.result()
 
             start_log_group(f"Performing checkup on {file}")
-            sys.stdout.write(log)
+            sys.stdout.write(logs)
             end_log_group()
 
             format_advice_map[file] = advice
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -3,13 +3,12 @@ import json
 from pathlib import Path, PurePath
 import subprocess
 import sys
-from tempfile import TemporaryDirectory
 from textwrap import indent
 from typing import Optional, List, Dict, Tuple
 import shutil
 
 from ..common_fs import FileObj
-from ..loggers import start_log_group, end_log_group, worker_log_file_init, logger
+from ..loggers import start_log_group, end_log_group, worker_log_init, logger
 from .clang_tidy import run_clang_tidy, TidyAdvice
 from .clang_format import run_clang_format, FormatAdvice
 
@@ -36,7 +35,6 @@ def assemble_version_exec(tool_name: str, specified_version: str) -> Optional[st
 
 def _run_on_single_file(
     file: FileObj,
-    temp_dir: str,
     log_lvl: int,
     tidy_cmd,
     checks,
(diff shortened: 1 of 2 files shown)


cpp_linter/clang_tools/__init__.py
```diff
@@ -157,10 +154,10 @@ def show_tool_version_output(cmd: str):  # show version output for executable us
         format_advice_map: Dict[str, Optional[FormatAdvice]] = {}
         tidy_notes_map: Dict[str, Optional[TidyAdvice]] = {}
         for future in as_completed(futures):
-            file, log_file, note, advice = future.result()
+            file, log, note, advice = future.result()
```
2bndy5, line 157:
```suggestion
            file, logs, note, advice = future.result()
```

cpp_linter/clang_tools/__init__.py
```diff
@@ -157,10 +154,10 @@ def show_tool_version_output(cmd: str):  # show version output for executable us
...
 
             start_log_group(f"Performing checkup on {file}")
-            sys.stdout.write(Path(log_file).read_text())
+            sys.stdout.write(log)
```
2bndy5, line 160:
```suggestion
            sys.stdout.write(logs)
```

2bndy5 (changes requested):
Sorry for the delayed response. I got lost exploring the `multiprocessing` docs and the `concurrent.futures` docs...

2bndy5:
Initial benchmark indeed looks a bit faster:
```
hyperfine "cpp-linter -l false -a false -j 1" "cpp-linter -l false -a false -j 0"
Benchmark 1: cpp-linter -l false -a false -j 1
  Time (mean ± σ):     54.647 s ±  2.467 s    [User: 20.625 s, System: 10.809 s]
  Range (min … max):   52.247 s … 57.656 s    5 runs

Benchmark 2: cpp-linter -l false -a false -j 0
  Time (mean ± σ):     27.898 s ±  0.882 s    [User: 24.238 s, System: 10.881 s]
  Range (min … max):   26.847 s … 28.713 s    5 runs

Summary
  cpp-linter -l false -a false -j 0 ran
    1.96 ± 0.11 times faster than cpp-linter -l false -a false -j 1
``` 
The previous PR averaged about 1.95 times faster with `-j 0` 🎉

2bndy5 (approved):

2bndy5:
> The behavior might cause memory problems on repositories with many files and warnings.

Its good that this was done in a separate PR. If we do get problems with memory saturation, then we can simply revert the merge commit and rollback to the temp file approach.

PR #103, opened 2024-04-24
Glob ignores

2bndy5:
satisfy ideas stated in cpp-linter/cpp-linter-action#233

This basically allows better reuse of code for filtering files. We can now filter all files (globally) and further filter files based on tool-specific patterns given by the new `--ignore-tidy` and/or `--ignore-format` options.

Any generated advice from clang tools is now stored in `FileObj` attributes. This allows better handling of advice _per file_. Previously, clang-tidy and clang-format were expected to run on all files, but this is not guaranteed now.

The collection of parsed CLI args is passed around instead of passing only relevant argument values. This keeps
- [x] functions' parameter counts down (less complexity)
- [x] functions signatures easier to maintain
- [x] less duplicated doc strings

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,6 @@ repos:
           - types-docutils
           - rich
           - pytest
-          - meson
           - requests-mock
           - '.'
   - repo: https://github.com/codespell-project/codespell
--- a/cpp_linter/__init__.py
+++ b/cpp_linter/__init__.py
@@ -3,10 +3,11 @@ If executed from command-line, then `main()` is the entrypoint.
 """
 
 import os
-from .common_fs import list_source_files, CACHE_PATH
+from .common_fs import CACHE_PATH
+from .common_fs.file_filter import FileFilter
 from .loggers import start_log_group, end_log_group, logger
 from .clang_tools import capture_clang_tools_output
-from .cli import cli_arg_parser, parse_ignore_option
+from .cli import cli_arg_parser
 from .rest_api.github_api import GithubApiClient
 
 
@@ -28,30 +29,34 @@ def main():
     logger.setLevel(10 if args.verbosity or rest_api_client.debug_enabled else 20)
 
     # prepare ignored paths list
-    ignored, not_ignored = parse_ignore_option(args.ignore, args.files)
+    global_file_filter = FileFilter(
+        extensions=args.extensions, ignore_value=args.ignore, not_ignored=args.files
+    )
+    global_file_filter.parse_submodules()
 
     # change working directory
     os.chdir(args.repo_root)
     CACHE_PATH.mkdir(exist_ok=True)
 
+    start_log_group("Get list of specified source files")
     if args.files_changed_only:
         files = rest_api_client.get_list_of_changed_files(
             extensions=args.extensions,
-            ignored=ignored,
-            not_ignored=not_ignored,
+            ignored=global_file_filter.ignored,
+            not_ignored=global_file_filter.not_ignored,
             lines_changed_only=args.lines_changed_only,
         )
         rest_api_client.verify_files_are_present(files)
     else:
-        files = list_source_files(args.extensions, ignored, not_ignored)
+        files = global_file_filter.list_source_files()
         # at this point, files have no info about git changes.
         # for PR reviews, we need this info
         if is_pr_event and (args.tidy_review or args.format_review):
             # get file changes from diff
             git_changes = rest_api_client.get_list_of_changed_files(
                 extensions=args.extensions,
-                ignored=ignored,
-                not_ignored=not_ignored,
+                ignored=global_file_filter.ignored,
+                not_ignored=global_file_filter.not_ignored,
                 lines_changed_only=0,  # prevent filtering out unchanged files
             )
             # merge info from git changes into list of all files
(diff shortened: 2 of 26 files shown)


cpp_linter/__init__.py
```diff
@@ -89,6 +90,9 @@ def main():
         tidy_review=is_pr_event and args.tidy_review,
         format_review=is_pr_event and args.format_review,
         num_workers=args.jobs,
+        extensions=args.extensions,
+        tidy_ignore=args.ignore_tidy,
+        format_ignore=args.ignore_format,
```
2bndy5, line 95: Most of these args are just passing directly from the `args` namespace created from `parse_args()`. It might be more efficient to pass either <sup>1</sup>a derivative of the CLI args namespace that is specifically applied to tool analysis or <sup>2</sup>the entire `args` object.

cpp_linter/cli.py
```diff
@@ -149,8 +145,31 @@
...
+      All asterisk characters (``*``) are not literal
+      as they were before. See
+      :py:meth:`~pathlib.Path.glob()` for more details
+      about Unix style glob patterns.
+""",
+)
+cli_arg_parser.add_argument(
+    "-M",
+    "--ignore-format",
+    default="",
```
2bndy5, line 161: I think the `FileFilter` class could be used as a `type` in the arg_parse API.

2bndy5: I opted away from this.

tests/test_cli_args.py
```diff
@@ -2,53 +2,12 @@
...
-    jobs: int = 1
+from cpp_linter.cli import cli_arg_parser, Args
 
 
 def test_defaults():
     """test default values"""
-    args = cli_arg_parser.parse_args("")
+    args = cli_arg_parser.parse_args("", namespace=Args)
     for key in args.__dict__.keys():
         assert args.__dict__[key] == getattr(Args, key)
```
2bndy5, line 12: I think this assert is now redundant. It basically says "is a == a?"

2bndy5: removed

cpp_linter/common_fs/file_filter.py
```diff
@@ -0,0 +1,216 @@
...
+            submodules.read(git_modules.resolve().as_posix())
+            for module in submodules.sections():
+                sub_mod_path = submodules[module]["path"]
+                if not self.is_file_in_list(
+                    ignored=False, file_name=PurePath(sub_mod_path)
+                ):
+                    logger.info(
+                        "Appending submodule to ignored paths: %s", sub_mod_path
+                    )
+                    self.ignored.add(sub_mod_path)
```
2bndy5, line 52: The submodule ignored path should be relative to the .gitmodules file's location. Currently this function only works well at repo-root.

cpp_linter/common_fs/file_filter.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def __init__(
+        self,
+        ignore_value: str = "",
+        extensions: Optional[List[str]] = None,
+        not_ignored: Optional[List[str]] = None,
+        tool_specific_name: Optional[str] = None,
+    ) -> None:
+        #: A set of file extensions that are considered C/C++ sources.
+        self.extensions = set(extensions or [])
```
2bndy5, line 28:
```suggestion
        self.extensions: Set[str] = set(extensions or [])
```

cpp_linter/common_fs/file_filter.py
```diff
@@ -0,0 +1,216 @@
...
+        """Parse a given string of paths (separated by a ``|``) into ``ignored`` and
+        ``not_ignored`` lists of strings.
+
+        :param paths: This argument conforms to the input value of :doc:`:doc:`CLI <cli_args>` <cli_args>` arg
+            :std:option:`--ignore`.
+
+        Results are added accordingly to the `ignored` and `not_ignored` attributes.
+        """
+        if paths:
+            for path in paths.split("|"):
```
2bndy5, line 64: I have to make sure that `|` is not a part of the Unix glob style syntax. If it is, then we'll have to change the pattern here (`|`) to something that isn't part of the glob style syntax (maybe like `;|`).

2bndy5: - https://en.wikipedia.org/wiki/Glob_(programming)
- https://man7.org/linux/man-pages/man7/glob.7.html

Both don't mention `|` as part of the Unix style glob syntax. I think we're good to go here?

.pre-commit-config.yaml
```diff
@@ -30,7 +30,6 @@ repos:
           - types-docutils
           - rich
           - pytest
-          - meson
```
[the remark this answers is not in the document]
2bndy5: Here it was on used for typing only.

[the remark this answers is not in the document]
2bndy5: It is still in the requirements-dev.txt
https://github.com/cpp-linter/cpp-linter/blob/glob-ignores/requirements-dev.txt#L2

2bndy5:
I know this one got a little out of control. One small refactor led to another and another... With the goal of cpp-linter/cpp-linter-action#233 accomplished, I'm ready to do a second round of testing and then begin merging this upstream.

PR #104, opened 2024-04-24
abstract `api_request()` with custom rate-limit headers

2bndy5:
also moves `has_more_pages()` into rest_api_client.py module.

This is a small abstraction from trying to support gitlab CI (see [gitlab PR](https://gitlab.com/2bndy5/cpp-linter/-/merge_requests/7/diffs)). It posses no significant API changes.

### DIFF
--- a/cpp_linter/rest_api/__init__.py
+++ b/cpp_linter/rest_api/__init__.py
@@ -1,8 +1,12 @@
+"""This base module holds abstractions common to using REST API.
+See other modules in ``rest_api`` subpackage for detailed derivatives.
+"""
+
 from abc import ABC
 from pathlib import PurePath
 import sys
 import time
-from typing import Optional, Dict, List, Any, cast
+from typing import Optional, Dict, List, Any, cast, NamedTuple
 import requests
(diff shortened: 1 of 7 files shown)


cpp_linter/rest_api/__init__.py
```diff
@@ -15,10 +17,46 @@
...
+    """A collection of HTTP response header keys that describe a REST API's rate limits.
+    Each parameter corresponds to a instance attribute (see below)."""
+
+    def __init__(self, reset: str, remaining: str, retry: str) -> None:
+        #: The header key of the rate limit's reset time.
+        self.reset = reset
+        #: The header key of the rate limit's remaining attempts.
+        self.remaining = remaining
+        #: The header key of the rate limit's "backoff" time interval.
+        self.retry = retry
```
2bndy5, line 30: This looks fine for docs, but I wish this would be a [named tuple](https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields) instead of a mutable object.

2bndy5:
Probably easier if this is merged first before #103

PR #107, opened 2024-06-01
allow PR reviews to be passive

2bndy5:
ref cpp-linter/cpp-linter-action#243


2bndy5:
This works as expected in the test repo.

PR #115, opened 2024-09-04
Switch to unidiff

2bndy5:
This migrates diff-parsing fallback mechanism (previously using regex patterns) to new dependecy [unidiff](https://pypi.org/project/unidiff).

This allows for a few new tactics:
- able to parse partial patches provided by REST API endpoints that list files with individual patch for each file (addresses cpp-linter/cpp-linter-action#249)
- able to move PR review comment generation from `rest_api/github_api.py` module into new `clang_tools.patcher.py` module because we don't need to keep data (previously used by `pygit2.Patch`) alive for the entire process (which avoided data corruption when the string of changes dropped out of scope).

   This change employs a mixin class to generate PR suggestions (for better/clearer code reuse).

I originally intended to use this to merge suggestions from both tools, but that isn't as easy as it sounds. Perhaps later, I can experiment with diff merging using these changes.


2bndy5:
It seems that our pr-review test wasn't generating suggestions with clang tools v7 and v8. I'm inclined to skip those versions when running tests, but It might be better to adjust the checks so that those versions generate suggestions.

Also, there seems to be a problem in CI where the coverage data (`.coverage*` files) isn't generated into repo root. This doesn't happen locally, so I'll have to investigate more in depth.

2bndy5:
> It seems that our pr-review test wasn't generating suggestions with clang tools v7 and v8

This was only true for the demo.hpp when using clang-tidy v7 or v8. It no longer makes sense to assert that suggestions would exists if the file was not changed by clang-tidy (or clang-format). I've opted to log a debug message for this scenario when generating PR review suggestions.

2bndy5:
> Also, there seems to be a problem in CI where the coverage data (`.coverage*` files) isn't generated into repo root. This doesn't happen locally, so I'll have to investigate more in depth.

This was a problem with `actions/upload-artifact`'s new input `include-hidden-files` which defaults to `false`. Setting this to `true` fixed the uploading of coverage data as artifacts.

cpp_linter/clang_tools/patcher.py
```diff
@@ -0,0 +1,67 @@
...
+        file_obj: FileObj,
+        summary_only: bool,
+    ):
+        """Create a list of suggestions from the tool's `patched` output."""
+
+        if not self.patched:
+            logger.debug(
+                "%s has no suggestions for %s", self.__class__.__name__, file_obj.name
+            )
+            return ("", [], 0)
```
2bndy5, line 35: This is no longer an assertion because the generated diff of suggested fixes is saved to `TidyAdvice.patched` or `FaormatAdvice.patched`. If there is no suggested fixes, then [`difflib.unified_diff()`](https://docs.python.org/3/library/difflib.html#difflib.unified_diff) returns an empty string. See also changes in [clang-tidy.py](https://github.com/cpp-linter/cpp-linter/pull/115/files#diff-f633f1dfb198669ddc2086ce6ac1e9817803a4ed775e5065d95f74bcd08d6f80) and [clang-format.py](https://github.com/cpp-linter/cpp-linter/pull/115/files#diff-5310aec821f0c94a8f355ea0a3e8fb389aa78764a93f8f483b8a3649defcc88f).

tests/demo/.clang-tidy
```diff
@@ -2,7 +2,6 @@
 Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-*,bugprone-*,clang-analyzer-*,mpi-*,misc-*,readability-*'
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
-AnalyzeTemporaryDtors: false
```
2bndy5, line 5: was incompatible with clang-tidy v18

.github/workflows/run-dev-tests.yml
```diff
@@ -102,6 +102,7 @@ jobs:
         with:
           name: coverage-data-${{ runner.os }}-py${{ matrix.py }}-${{ matrix.version }}
           path: .coverage*
+          include-hidden-files: true
```
2bndy5, line 105: > Also, there seems to be a problem in CI where the coverage data (`.coverage*` files) isn't generated into repo root. This doesn't happen locally, so I'll have to investigate more in depth.

This was a problem with `actions/upload-artifact`'s new input `include-hidden-files` which defaults to `false`. Setting this to `true` fixed the uploading of coverage data as artifacts.

cpp_linter/rest_api/github_api.py
```diff
@@ -111,12 +102,61 @@ def get_list_of_changed_files(
...
+                                [hunk_start, hunk_start + hunk.target_length]
+                            )
+                            for line in hunk:
+                                if line.line_type == "+":
+                                    line_number = line.target_line_no
+                                    assert line_number is not None
+                                    additions.append(line_number)
+                        if has_line_changes(lines_changed_only, diff_chunks, additions):
+                            files.append(FileObj(file_name, additions, diff_chunks))
+                return files
```
2bndy5, line 157: I should abstract these 50 lines into a new private function.

cpp_linter/rest_api/github_api.py
```diff
@@ -432,54 +472,12 @@ def create_review_comments(
...
-                else:
-                    body += f"\n```suggestion\n{suggestion}```"
-                comment["body"] = body
-                comments.append(comment)
+            full_patch += patch
+            total += hunk_total
+            comments.extend(suggestions)
 
             # now check for clang-tidy warnings with no fixes applied
             if tidy_tool and file_obj.tidy_advice:
```
2bndy5, line 483: This could be moved into `TidyAdvice.get_suggestions_from_patch()`.

cpp_linter/clang_tools/patcher.py
```diff
@@ -0,0 +1,67 @@
...
+                    assert line_numb is not None
+                    removed.append(line_numb)
+            if not suggestion and removed:
+                body += "\nPlease remove the line(s)\n- "
+                body += "\n- ".join([str(x) for x in removed])
+            else:
+                body += f"\n```suggestion\n{suggestion}```"
+            comment["body"] = body
+            comments.append(comment)
+        return (self.patched, comments, total)
```
2bndy5, line 67: Currently, this returns `comments` in a dict that is only compatible with GitHub REST API. I think it would be better to return a data struct that can be converted into something compatible with the git server's REST API that's being used.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pygit2
```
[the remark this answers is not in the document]
2bndy5: Nope. I'm still using pygit2 for parsing full diffs, getting diffs from a local env (no CI), and driving a unit test about assesing staged vs unstaged changes.

But I'm now using unidiff to parse partial diffs and suggested fixes. This combination should be more reliable than only using pygit2

cpp_linter/rest_api/github_api.py
2bndy5: done

cpp_linter/clang_tools/patcher.py
2bndy5: done.

I implemented an approach similar to what I plan to use in the rust port.

cpp_linter/rest_api/github_api.py
2bndy5: done

2bndy5:
uh oh. The generated diff from `diff_lib.unified_diff()` does not handle similarities very well (not at all actually). See [test repo's PR review](https://github.com/cpp-linter/test-cpp-linter-action/pull/39#pullrequestreview-2281822548).

I'm going to have to rethink and revert somethings here.

2bndy5:
Ok, `diff_lib.unified_diff()` kind of works when `--lines-changed-only` is true, but I think it expects the line endings to be LF. If the lines end with CRLF, then the whole file gets adjusted to use LF endings (by clang tools according to the specified style guide), and the generated diff encompasses the entire file.

2bndy5:
I'm refactoring the branch and will open a new PR shortly.

PR #116, opened 2024-09-05
resort to paginated request(s) to get file changes

2bndy5:
This uses the paginated JSON responses from Github's REST API to get file changes (with individual diffs) when the full diff is too large for 1 HTTP request/response.
- [endpoint that lists the files in a PR](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files)
- [endpoint that lists the files in a commits](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit)

These changes are abstracted from the initial proposal in #115 and solely aims to address cpp-linter/cpp-linter-action#249

### DIFF
--- a/.github/workflows/run-dev-tests.yml
+++ b/.github/workflows/run-dev-tests.yml
@@ -102,6 +102,7 @@ jobs:
         with:
           name: coverage-data-${{ runner.os }}-py${{ matrix.py }}-${{ matrix.version }}
           path: .coverage*
+          include-hidden-files: true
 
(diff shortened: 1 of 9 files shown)


2bndy5:
With the new unit tests (modeled from actual HTTP responses), I'm confident this is safe to merge. It would be too much work to create a test condition where we have to parse over 300 file changes just to satisfy these changes.

PR #121, opened 2024-09-18
swap file names in paginated changed files's diff

2bndy5:
A bug-fix about changes in #116 

When getting changed files for a large diff, I noticed the generated diff from the files' individual patch had the filenames reversed. This only affects changed files that were renamed.


2bndy5:
The codecov site has been buggy today. There is no summary/response in cpp-linter/cpp_linter_rs#37 PR either. I'll wait until tomorrow. Hopefully, it'll be back in working order.

PS - I discovered this when porting the feature to rust. I'm surprised that the unit tests didn't pick this up.

PR #131, opened 2024-12-05
Hide review comments and resolve or delete review conversations

maoliver-amd:
This extends the existing cang-tidy/format review capability by adding additional functionality:

- Auto hide/minimize old outdated review comments
- Mark outdated review conversations as resolved
- Delete outdated review conversations (user option to select between resolving or deleting)
- Reuse existing review conversations on subsequent runs in case issue is still present (prevent spam comments for identical issues being continuously added on each run)(user option to either resue or resolve/delete and repost)
- Reuse existing review comment on subsequent runs for issues still present while also creating a new comment should any additional new issues be detected (reduces duplicate issue spam)

Uses graphql queries to provide additional functionality beyond what the github api suppots

## Release Notes

- **New Features**
  - Introduced options to manage pull request review comments: delete outdated comments and reuse existing comments.
  - Enhanced GitHub API client with new GraphQL queries and mutations for improved review comment management.
  - Added functionality to track and report reused suggestions in review summaries.

- **Bug Fixes**
  - Improved handling of existing review comments, including fetching, closing, and hiding stale reviews.

- **Tests**
  - Added mock GraphQL request handling in tests to simulate interactions with the GitHub API for review comments.

### DIFF
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -70,9 +70,9 @@ class Args(UserDict):
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
     #: See :std:option:`--delete-review-comments`.
-    delete_comments: bool = False
+    delete_review_comments: bool = False
     #: See :std:option:`--reuse-review-comments`.
-    reuse_comments: bool = True
+    reuse_review_comments: bool = True
 
 
 _parser_args: Dict[Sequence[str], Any] = {}
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -362,19 +362,23 @@ _parser_args[("-R", "--passive-reviews")] = dict(
     help="""Set to ``true`` to prevent Pull Request
 reviews from requesting or approving changes.""",
 )
-_parser_args[("-C", "--delete-comments")] = dict(
+_parser_args[("-C", "--delete-review-comments")] = dict(
     default="true",
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to delete existing outdated/unused
-review comments, ``false`` to just set them to resolved.
+Pull request review comments, ``false`` to just set them to resolved.
+This only effects review comments made when either
+:std:option:`--tidy-review` or :std:option:`--format-review` is enabled.
 
 Defaults to ``%(default)s``.""",
 )
-_parser_args[("-U", "--reuse-comments")] = dict(
+_parser_args[("-U", "--reuse-review-comments")] = dict(
     default="true",
     type=lambda input: input.lower() == "true",
-    help="""Set to ``true`` to reuse existing review
-comments if nothing has changed instead of making new ones.
+    help="""Set to ``true`` to reuse existing Pull request
+review comments if nothing has changed instead of making new ones.
+This only effects review comments made when either
+:std:option:`--tidy-review` or :std:option:`--format-review` is enabled.
 
 Defaults to ``%(default)s``.""",
 )
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -70,7 +70,7 @@ class Args(UserDict):
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
     #: See :std:option:`--delete-review-comments`.
-    delete_comments: bool = True
+    delete_comments: bool = False
     #: See :std:option:`--reuse-review-comments`.
     reuse_comments: bool = True
 
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -69,6 +69,10 @@ class Args(UserDict):
     ignore_format: str = ""
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
+    #: See :std:option:`--delete-comments`.
+    delete_comments: bool = True
+    #: See :std:option:`--reuse-comments`.
+    reuse_comments: bool = True
 
 
 _parser_args: Dict[Sequence[str], Any] = {}
@@ -357,6 +361,22 @@ _parser_args[("-R", "--passive-reviews")] = dict(
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to prevent Pull Request
 reviews from requesting or approving changes.""",
+)
+_parser_args[("-C", "--delete-comments")] = dict(
+    default="true",
+    type=lambda input: input.lower() == "true",
+    help="""Set to ``true`` to delete existing outdated/unused
+review comments, ``false`` to just set them to resolved.
+
+Defaults to ``%(default)s``.""",
+)
+_parser_args[("-U", "--reuse-comments")] = dict(
+    default="true",
+    type=lambda input: input.lower() == "true",
+    help="""Set to ``true`` to reuse existing review
+comments if nothing has changed instead of making new ones.
+
+Defaults to ``%(default)s``.""",
 )
 
 
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -69,9 +69,9 @@ class Args(UserDict):
     ignore_format: str = ""
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
-    #: See :std:option:`--delete-comments`.
+    #: See :std:option:`--delete-review-comments`.
     delete_comments: bool = True
-    #: See :std:option:`--reuse-comments`.
+    #: See :std:option:`--reuse-review-comments`.
     reuse_comments: bool = True
 
 
@@ -363,7 +363,7 @@ _parser_args[("-R", "--passive-reviews")] = dict(
 reviews from requesting or approving changes.""",
 )
 _parser_args[("-C", "--delete-review-comments")] = dict(
-    default="true",
+    default="false",
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to delete existing outdated/unused
 Pull request review comments, ``false`` to just set them to resolved.
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -71,8 +71,6 @@ class Args(UserDict):
     passive_reviews: bool = False
     #: See :std:option:`--delete-review-comments`.
     delete_review_comments: bool = False
-    #: See :std:option:`--reuse-review-comments`.
-    reuse_review_comments: bool = True
 
 
 _parser_args: Dict[Sequence[str], Any] = {}
@@ -367,16 +365,6 @@ _parser_args[("-C", "--delete-review-comments")] = dict(
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to delete existing outdated/unused
 Pull request review comments, ``false`` to just set them to resolved.
-This only effects review comments made when either
-:std:option:`--tidy-review` or :std:option:`--format-review` is enabled.
-
-Defaults to ``%(default)s``.""",
-)
-_parser_args[("-U", "--reuse-review-comments")] = dict(
-    default="true",
-    type=lambda input: input.lower() == "true",
-    help="""Set to ``true`` to reuse existing Pull request
-review comments if nothing has changed instead of making new ones.
 This only effects review comments made when either
 :std:option:`--tidy-review` or :std:option:`--format-review` is enabled.
 
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -365,7 +365,7 @@ _parser_args[("-C", "--delete-review-comments")] = dict(
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to delete existing outdated/unused
 Pull request review comments, ``false`` to just set them to resolved.
-This only effects review comments made when either
+This only affects review comments made when either
 :std:option:`--tidy-review` or :std:option:`--format-review` is enabled.
 
 Defaults to ``%(default)s``.""",
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -620,7 +620,7 @@ class GithubApiClient(RestApiClient):
             self.pull_request,
         )
         response = self.api_request(
-            url="https://api.github.com/graphql",
+            url=f"{self.api_url}/graphql",
             method="POST",
             data=json.dumps(
                 {"query": query}
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -342,8 +342,8 @@ class GithubApiClient(RestApiClient):
                 no_lgtm=args.no_lgtm,
                 passive_reviews=args.passive_reviews,
                 clang_versions=clang_versions,
-                delete_comments=args.delete_comments,
-                reuse_comments=args.reuse_comments,
+                delete_review_comments=args.delete_review_comments,
+                reuse_review_comments=args.reuse_review_comments,
             )
 
     def make_annotations(
@@ -480,8 +480,8 @@ class GithubApiClient(RestApiClient):
         no_lgtm: bool,
         passive_reviews: bool,
         clang_versions: ClangVersions,
-        delete_comments: bool = True,
-        reuse_comments: bool = True,
+        delete_review_comments: bool = True,
+        reuse_review_comments: bool = True,
     ):
         url = f"{self.api_url}/repos/{self.repo}/pulls/{self.pull_request}"
         response = self.api_request(url=url)
@@ -516,10 +516,10 @@ class GithubApiClient(RestApiClient):
         ignored_reviews = []
         if not summary_only:
             found_threads = self._get_existing_review_comments(
-                no_dismissed=reuse_comments and not delete_comments
+                no_dismissed=reuse_review_comments and not delete_review_comments
             )
             if found_threads:
-                if reuse_comments:
+                if reuse_review_comments:
                     # Keep already posted comments if they match new ones
                     review_comments_suggestions = review_comments.suggestions
                     review_comments.suggestions = []
@@ -558,7 +558,7 @@ class GithubApiClient(RestApiClient):
                                     break
                             if not found:
                                 self._close_review_comment(
-                                    thread["id"], comment["id"], delete_comments
+                                    thread["id"], comment["id"], delete_review_comments
                                 )
                     for suggestion in review_comments_suggestions:
                         if suggestion not in existing_review_comments:
@@ -568,7 +568,7 @@ class GithubApiClient(RestApiClient):
                     for thread in found_threads:
                         for comment in thread["comments"]["nodes"]:
                             self._close_review_comment(
-                                thread["id"], comment["id"], delete_comments
+                                thread["id"], comment["id"], delete_review_comments
                             )
         self._hide_stale_reviews(ignored_reviews=ignored_reviews)
         if len(review_comments.suggestions) == 0 and len(ignored_reviews) > 0:
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -528,9 +528,8 @@ class GithubApiClient(RestApiClient):
                     for thread in found_threads:
                         for comment in thread["comments"]["nodes"]:
                             found = False
-                            assert (
-                                "originalLine" in comment
-                            ), "GraphQL response malformed"
+                            if "originalLine" not in comment:
+                                raise KeyError("GraphQL response malformed: 'originalLine' missing in comment")
                             line_start = comment.get("startLine", None) or comment.get("originalStartLine", None) or -1
                             line_end = comment.get("line", None) or comment["originalLine"]
                             for suggestion in review_comments_suggestions:
@@ -691,11 +690,7 @@ class GithubApiClient(RestApiClient):
                             and thread["isCollapsed"] is False
                         )
                     )
-                    and comment["author"]["login"] == "github-actions"
-                    and (
-                        comment["body"].strip().startswith("### clang-format")
-                        or comment["body"].strip().startswith("### clang-tidy")
-                    )
+                    and comment["body"].startswith(COMMENT_MARKER)
                 ):
                     found_threads.append(thread)
                     break
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -418,7 +418,7 @@ class GithubApiClient(RestApiClient):
         passive_reviews: bool,
         clang_versions: ClangVersions,
         delete_comments: bool = True,
-        reuse_comments: bool = True
+        reuse_comments: bool = True,
     ):
         url = f"{self.api_url}/repos/{self.repo}/pulls/{self.pull_request}"
         response = self.api_request(url=url)
@@ -453,7 +453,8 @@ class GithubApiClient(RestApiClient):
         ignored_reviews = []
         if not summary_only:
             found_threads = self._get_existing_review_comments(
-                no_dismissed=reuse_comments and not delete_comments)
+                no_dismissed=reuse_comments and not delete_comments
+            )
             if found_threads:
                 if reuse_comments:
                     # Keep already posted comments if they match new ones
@@ -471,17 +472,31 @@ class GithubApiClient(RestApiClient):
                                 )
                                 line_end = comment.get("line", line_end)
                             for suggestion in review_comments_suggestions:
-                                if (suggestion.file_name == comment["path"] and suggestion.line_start == line_start and suggestion.line_end == line_end and suggestion.comment == comment["body"] and suggestion not in existing_review_comments and thread["isResolved"] is False and thread["isCollapsed"] is False):
+                                if (
+                                    suggestion.file_name == comment["path"]
+                                    and suggestion.line_start == line_start
+                                    and suggestion.line_end == line_end
+                                    and suggestion.comment == comment["body"]
+                                    and suggestion not in existing_review_comments
+                                    and thread["isResolved"] is False
+                                    and thread["isCollapsed"] is False
+                                ):
                                     found = True
                                     logger.info(
-                                        "Using existing review comment: path='%s', line_start='%s', line_end='%s'", comment["path"], line_start, line_end)
+                                        "Using existing review comment: path='%s', line_start='%s', line_end='%s'",
+                                        comment["path"],
+                                        line_start,
+                                        line_end,
+                                    )
                                     ignored_reviews.append(
-                                        comment["pullRequestReview"]["id"])
+                                        comment["pullRequestReview"]["id"]
+                                    )
                                     existing_review_comments.append(suggestion)
                                     break
-                            if (not found):
+                            if not found:
                                 self._close_review_comment(
-                                    thread["id"], comment["id"], delete_comments)
+                                    thread["id"], comment["id"], delete_comments
+                                )
                     for suggestion in review_comments_suggestions:
                         if suggestion not in existing_review_comments:
                             review_comments.suggestions.append(suggestion)
@@ -490,7 +505,8 @@ class GithubApiClient(RestApiClient):
                     for thread in found_threads:
                         for comment in thread["comments"]["nodes"]:
                             self._close_review_comment(
-                                thread["id"], comment["id"], delete_comments)
+                                thread["id"], comment["id"], delete_comments
+                            )
         self._hide_stale_reviews(ignored_reviews=ignored_reviews)
         if len(review_comments.suggestions) == 0 and len(ignored_reviews) > 0:
             logger.info("Using previous review as nothing new was found")
@@ -623,14 +639,13 @@ class GithubApiClient(RestApiClient):
         response = self.api_request(
             url=f"{self.api_url}/graphql",
             method="POST",
-            data=json.dumps(
-                {"query": query}
-            ),
-            strict=False
+            data=json.dumps({"query": query}),
+            strict=False,
         )
         if response.status_code != 200:
             logger.error(
-                "Could not get existing review comments: %d", response.status_code)
+                "Could not get existing review comments: %d", response.status_code
+            )
             return
         data = response.json()
         found_threads = []
@@ -638,15 +653,26 @@ class GithubApiClient(RestApiClient):
             for comment in thread["comments"]["nodes"]:
                 if (
                     comment["id"]
-                    and (not no_dismissed or (thread["isResolved"] is False and thread["isCollapsed"] is False))
+                    and (
+                        not no_dismissed
+                        or (
+                            thread["isResolved"] is False
+                            and thread["isCollapsed"] is False
+                        )
+                    )
                     and comment["author"]["login"] == "github-actions"
-                    and (comment["body"].strip().startswith("### clang-format") or comment["body"].strip().startswith("### clang-tidy"))
+                    and (
+                        comment["body"].strip().startswith("### clang-format")
+                        or comment["body"].strip().startswith("### clang-tidy")
+                    )
                 ):
                     found_threads.append(thread)
                     break
         return found_threads
 
-    def _close_review_comment(self, thread_id: str, comment_id: str, delete: bool = True):
+    def _close_review_comment(
+        self, thread_id: str, comment_id: str, delete: bool = True
+    ):
         """Resolve or Delete an existing review comment.
 
         :param thread_id: Thread ID for the conversation to close (only used when ``delete``==`False`).
@@ -661,9 +687,7 @@ class GithubApiClient(RestApiClient):
                 }
             }
         }
-        """ % (
-            thread_id
-        )
+        """ % (thread_id)
         if delete:
             mutation = """
             mutation {
@@ -673,25 +697,21 @@ class GithubApiClient(RestApiClient):
                     }
                 }
             }
-            """ % (
-                comment_id
-            )
+            """ % (comment_id)
         response = self.api_request(
             url="https://api.github.com/graphql",
             method="POST",
-            data=json.dumps(
-                {"query": mutation}
-            ),
-            strict=False
+            data=json.dumps({"query": mutation}),
+            strict=False,
         )
         if response.status_code != 200:
-            logger.error("Failed to close review comment: %d",
-                         response.status_code)
+            logger.error("Failed to close review comment: %d", response.status_code)
         elif "errors" in response.json():
             error_msg = response.json()["errors"][0]["message"]
             if "Resource not accessible by integration" in error_msg:
                 logger.error(
-                    "Closing review comments requires `contents: write` permission.")
+                    "Closing review comments requires `contents: write` permission."
+                )
             else:
                 logger.error("Closing review comment failed: %s", error_msg)
         else:
@@ -722,28 +742,24 @@ class GithubApiClient(RestApiClient):
                             }
                         }
                     }
-                    """ % (
-                        review["node_id"]
-                    )
+                    """ % (review["node_id"])
                     response = self.api_request(
                         url="https://api.github.com/graphql",
                         method="POST",
-                        data=json.dumps(
-                            {"query": mutation}
-                        ),
-                        strict=False
+                        data=json.dumps({"query": mutation}),
+                        strict=False,
                     )
                     if response.status_code != 200:
                         logger.error(
-                            "Failed to hide review comment: %d", response.status_code)
+                            "Failed to hide review comment: %d", response.status_code
+                        )
                     elif "errors" in response.json():
                         error_msg = response.json()["errors"][0]["message"]
                         if "Resource not accessible by integration" in error_msg:
                             logger.error(
-                                "Hiding review comments requires `contents: write` permission.")
+                                "Hiding review comments requires `contents: write` permission."
+                            )
                         else:
-                            logger.error(
-                                "Hiding review comment failed: %s", error_msg)
+                            logger.error("Hiding review comment failed: %s", error_msg)
                     else:
-                        logger.debug(
-                            "Review comment minimized: %s", review["node_id"])
+                        logger.debug("Review comment minimized: %s", review["node_id"])
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -37,6 +37,69 @@ RATE_LIMIT_HEADERS = RateLimitHeaders(
     retry="retry-after",
 )
 
+QUERY_REVIEW_COMMENTS = """
+query {
+    repository(owner:"%s", name:"%s") {
+        pullRequest(number: %d) {
+            id
+            reviewThreads(last: 100) {
+                nodes {
+                    id
+                    isResolved
+                    isCollapsed
+                    comments(first: 10) {
+                        nodes {
+                            id
+                            body
+                            path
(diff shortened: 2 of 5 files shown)


cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +572,178 @@ def _dismiss_stale_reviews(self, url: str):
...
+                    break
+        return found_threads
+
+    def _close_review_comment(self, thread_id: str, comment_id: str, delete: bool = True):
+        """Resolve or Delete an existing review comment.
+
+        :param thread_id: Thread ID for the conversation to close (only used when ``delete``==`False`).
+        :param comment_id: The comment ID of the comment within the requested thread to close (only used when ``delete``==`True`).
+        :param delete: `True` to delete the review comment, `False` to set it as resolved.
+        """
```
maoliver-amd, line 655: Uses graphQL to either set a review comment conversation as resolved or to delete based on user options.

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +572,178 @@ def _dismiss_stale_reviews(self, url: str):
...
+            else:
+                logger.error("Closing review comment failed: %s", error_msg)
+        else:
+            logger.debug("Review comment closed: %s", thread_id)
+
+    def _hide_stale_reviews(self, ignored_reviews: List[str]):
+        """Hide all review comments that were previously created by cpp-linter
+
+        :param ignored_reviews: List of review comments to keep displayed.
+        """
```
maoliver-amd, line 704: Uses graphQL to hide old stale review comments from previous runs

cpp_linter/cli.py
```diff
@@ -358,6 +362,22 @@ class Args(UserDict):
...
 reviews from requesting or approving changes.""",
 )
+_parser_args[("-C", "--delete-comments")] = dict(
+    default="true",
+    type=lambda input: input.lower() == "true",
+    help="""Set to ``true`` to delete existing outdated/unused
+review comments, ``false`` to just set them to resolved.
+
+Defaults to ``%(default)s``.""",
+)
```
2bndy5, line 372: The behavior when `--thread-comments` is set to 'true'. Why do we need a separate/new CLI option?

cpp_linter/cli.py
```diff
@@ -358,6 +362,22 @@ class Args(UserDict):
...
+_parser_args[("-U", "--reuse-comments")] = dict(
+    default="true",
+    type=lambda input: input.lower() == "true",
+    help="""Set to ``true`` to reuse existing review
+comments if nothing has changed instead of making new ones.
+
+Defaults to ``%(default)s``.""",
+)
```
2bndy5, line 380: The behavior when `--thread-comments` is set to 'update'. Why do we need a separate/new CLI option?

2bndy5 (commented):
I suspected there was a way to do that, but I never checked the graphql endpoints. Thanks!

cpp_linter/cli.py
maoliver-amd: This is a terminology difference (perhaps '--delete-review-comments' or something)
`--thread-comments` effects these comments:
comment
whereas this option effects these comments:
image
Github differentiates them as the former are pull request comments and the latter are pull request review thread comments and get handled differently. The latter require the graphQL functionality in this PR to change

maoliver-amd: If the new options arent clear then feel free to suggest better naming

2bndy5: oh I see. Maybe a more explicit otion name then? like `--delete-review-comments` or `--reuse-review-comments`.

maoliver-amd: Yep thats probably a good suggestion, Ive update the options in this PR accordingly

cpp_linter/rest_api/github_api.py
```diff
@@ -446,6 +450,51 @@ def post_review(
...
+                    existing_review_comments = []
+                    for thread in found_threads:
+                        for comment in thread["comments"]["nodes"]:
+                            found = False
+                            line_start = comment["originalStartLine"] if comment["originalStartLine"] is not None else -1
+                            line_end = comment["originalLine"]
+                            if comment["startLine"] is not None:
+                                line_start = comment["startLine"] if comment["startLine"] is not None else (
+                                    comment["line"] if comment["line"] is not None else line_start)
+                                line_end = comment["line"] if comment["line"] is not None else line_end
```
2bndy5, line 471: I understand that after all that `... if ... else ...`, this is basically the same, yes?
```suggestion
                            if "line" in comment and comment["line"] is not None:
                                line_end = comment["line"]
                                line_start = line_end
                                if "startLine" in comment and comment["startLine"] is not None:
                                    line_start = comment["startLine"]
```

cpp_linter/rest_api/github_api.py
```diff
@@ -446,6 +450,51 @@ def post_review(
...
+                            for suggestion in review_comments_suggestions:
+                                if (suggestion.file_name == comment["path"] and suggestion.line_start == line_start and suggestion.line_end == line_end and suggestion.comment == comment["body"] and suggestion not in existing_review_comments and thread["isResolved"] is False and thread["isCollapsed"] is False):
```
2bndy5, line 473: Yikes! This is why we use ruff or black to format code:
```suggestion
                                if (
                                    suggestion.file_name == comment["path"]
                                    and suggestion.line_start == line_start
                                    and suggestion.line_end == line_end
                                    and suggestion.comment == comment["body"]
                                    and suggestion not in existing_review_comments
                                    and thread["isResolved"] is False
                                    and thread["isCollapsed"] is False
                                ):
```

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +571,178 @@ def _dismiss_stale_reviews(self, url: str):
...
                     )
+
+    def _get_existing_review_comments(self, no_dismissed: bool = True):
+        """Creates the list existing conversation threads to close.
+
+        :param no_dismissed: `True` to ignore any already dismissed comments.
+        """
+        repo_owner, repo_name = self.repo.split("/")
+        query = """
+        query {
```
2bndy5, line 586: I feel like these queries could be constants scoped to rest_api/github.py
```py
QUERY_REVIEW_COMMENTS = """ ... """
RESOLVE_REVIEW_COMMENT = """ ... """
DELETE_REVIEW_COMMENT = """ ... """
HIDE_REVIEW_COMMENT = """ ... """
```

cpp_linter/cli.py
```diff
@@ -358,6 +362,26 @@ class Args(UserDict):
     help="""Set to ``true`` to prevent Pull Request
 reviews from requesting or approving changes.""",
 )
+_parser_args[("-C", "--delete-review-comments")] = dict(
+    default="true",
```
2bndy5, line 366: ## Don't delete on default
```suggestion
    default="false",
```
I'd prefer to leave the more destructive operations as opt-in. My main concern here is any replies to the initial/deleted review comment(s) would become harder to understand.

2bndy5 (changes requested):
just at first look

2bndy5:
You'll also need to update the PR reviews test with mocked responses to the graphQL API. https://github.com/cpp-linter/cpp-linter/blob/baee3c2970a57a07ec22413aecbc0e2de8eccc7a/tests/reviews/test_pr_review.py#L119
You'll find more about that at [`requests-mock` lib's docs](https://requests-mock.readthedocs.io/en/latest/response.html#registering-responses)

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +571,178 @@ def _dismiss_stale_reviews(self, url: str):
...
+                }
+            }
+        }
+        """ % (
+            repo_owner,
+            repo_name,
+            self.pull_request,
+        )
+        response = self.api_request(
+            url="https://api.github.com/graphql",
```
2bndy5, line 623: I believe the URL `https://api.github.com/graphql` could also be abstracted from the REST API URL.
```suggestion
            url=f"{self.api_url}/graphql",
```
This abstraction makes testing with a mock server very easy.

cpp_linter/rest_api/github_api.py
maoliver-amd: No not quite. Theres a few combinations of things that can happen here. Firstly there are the original lines ("originalLine" and "originalStartLine") which are the defaults unless a comment has moved due to line changes in which case "startLine" and "line" are used. However single line comments dont have both as only multi line comments have both. So for instance `line_end` should be "line" if it exists, "line" is only valid if theres a "startLine", if neither then it should be "originalLine"
I did a bunch of tests on a repo with single and multi-line comments that were new/moved due to changes and the existing logic seemed to be the one that was correct for all those cases

cpp_linter/rest_api/github_api.py
maoliver-amd: I used pep8 for formatting, ive updated it again using ruff

cpp_linter/rest_api/github_api.py
maoliver-amd: done

cpp_linter/rest_api/github_api.py
2bndy5: we have pre-commit setup for this repo also.

cpp_linter/cli.py
```diff
@@ -69,6 +69,10 @@ class Args(UserDict):
     ignore_format: str = ""
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
+    #: See :std:option:`--delete-review-comments`.
+    delete_comments: bool = True
```
2bndy5, line 73:
```suggestion
    delete_comments: bool = False
```

Because we changed the default below

cpp_linter/rest_api/github_api.py
```diff
@@ -446,6 +513,68 @@ def post_review(
...
+                    for thread in found_threads:
+                        for comment in thread["comments"]["nodes"]:
+                            found = False
+                            line_start = comment.get("originalStartLine", -1)
+                            line_end = comment["originalLine"]
+                            if comment["startLine"] is not None:
+                                line_start = comment.get(
+                                    "startLine", comment.get("line", line_start)
+                                )
+                                line_end = comment.get("line", line_end)
```
2bndy5, line 536: I'm really just worried about `KeyError`s here.

If `"originalLine"` is guaraunteed to be in `comment`, then
```python
line_end = comment["originalLine"]
```
will not fail. Usually, we put `assert "originalLine" in comment` beforehand for typing assurance.

But still, you've got dead code here:
```python
                            if comment["startLine"] is not None:
                                line_start = comment.get(
                                    "startLine", comment.get("line", line_start)
                                )
```
If
```python
comment.get("startLine", None) == comment["startLine"]
```
then the code will be short-circuited to
```python
                            if comment.get("startLine", None) is not None:
                                line_start = comment["startLine"]
```

---
## Another swing
Am I getting close?
```suggestion
                            assert "originalLine" in comment, "GraphQL response malformed"
                            line_end = comment.get("line", comment.get("originalLine", -1))
                            line_start = comment.get("startLine", comment.get(
                                "line", comment.get("originalStartLine", line_end)
                            )
```

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +651,122 @@ def _dismiss_stale_reviews(self, url: str):
...
+
+        :param thread_id: Thread ID for the conversation to close (only used when ``delete``==`False`).
+        :param comment_id: The comment ID of the comment within the requested thread to close (only used when ``delete``==`True`).
+        :param delete: `True` to delete the review comment, `False` to set it as resolved.
+        """
+        mutation = RESOLVE_REVIEW_COMMENT % (thread_id)
+        if delete:
+            mutation = DELETE_REVIEW_COMMENT % (comment_id)
+        response = self.api_request(
+            url="https://api.github.com/graphql",
```
2bndy5, line 717:
```suggestion
            url=f"{self.api_url}/graphql",
```

2bndy5 (changes requested):

cpp_linter/rest_api/github_api.py
maoliver-amd: Hhere are some examples of the graphql response for various review thread comments that can be used to help determine the correct logic. All of these are actually taken from this PR at time of posting:
```
{
              "id": "PRRT_kwDOHzaZPM5HWwCo",
              "isResolved": false,
              "isCollapsed": false,
              "comments": {
                "nodes": [
                  {
                    "id": "PRRC_kwDOHzaZPM5vh3vq",
                    "body": "```suggestion\r\n                            line_start = comment.get(\"originalStartLine\", -1)\r\n```\r\nThere's a lot of in-lined conditionals here. Most of them seem like they could be replaced with similar suggestions. There might need to be some added typing.",
                    "path": "cpp_linter/rest_api/github_api.py",
                    "line": null,
                    "startLine": null,
                    "originalLine": 467,
                    "originalStartLine": null
                  }
                ]
              }
            },
            {
              "id": "PRRT_kwDOHzaZPM5HWxv0",
              "isResolved": true,
              "isCollapsed": true,
              "comments": {
                "nodes": [
                  {
                    "id": "PRRC_kwDOHzaZPM5vh6Wa",
                    "body": "```suggestion\r\n    #: See :std:option:`--delete-review-comments`.\r\n    delete_comments: bool = True\r\n    #: See :std:option:`--reuse-review-comments`.\r\n    reuse_comments: bool = True\r\n```",
                    "path": "cpp_linter/cli.py",
                    "line": null,
                    "startLine": null,
                    "originalLine": 75,
                    "originalStartLine": 72
                  }
                ]
              }
            },
            {
              "id": "PRRT_kwDOHzaZPM5HaLjj",
              "isResolved": false,
              "isCollapsed": false,
              "comments": {
                "nodes": [
                  {
                    "id": "PRRC_kwDOHzaZPM5vnTaI",
                    "body": "I'm really just worried about `KeyError`s here.\r\n\r\nIf `\"originalLine\"` is guaranteed to be in `comment`, then\r\n```python\r\nline_end = comment[\"originalLine\"]\r\n```\r\nwill not fail. Usually, we put `assert \"originalLine\" in comment` beforehand for typing assurance.\r\n\r\nBut still, you've got dead code here:\r\n```python\r\n                            if comment[\"startLine\"] is not None:\r\n                                line_start = comment.get(\r\n                                    \"startLine\", comment.get(\"line\", line_start)\r\n
                     )\r\n```\r\nIf\r\n```python\r\ncomment.get(\"startLine\", None) == comment[\"startLine\"]\r\n```\r\nthen the code will be short-circuited to\r\n```python\r\n                            if comment.get(\"startLine\", None) is not None:\r\n                                line_start = comment[\"startLine\"]\r\n```\r\n\r\n---\r\n## Another swing\r\nAm I getting close?\r\n```suggestion\r\n                            assert \"originalLine\" in comment, \"GraphQL response malformed\"\r\n                            line_end = comment.get(\"line\", comment[\"originalLine\"])\r\n                            line_start = comment.get(\"originalStartLine\", line_end)\r\n                            if \"line\" in comment:\r\n                                line_start = comment.get(\"startLine\", comment[\"line\"])\r\n```",
                    "path": "cpp_linter/rest_api/github_api.py",
                    "line": 536,
                    "startLine": 530,
                    "originalLine": 536,
                    "originalStartLine": 530
                  }
                ]
              }
            },
```
The only example we dont have from this thread is where we have something along these lines:
```
                    "line": 465,
                    "startLine": null,
                    "originalLine": 467,
                    "originalStartLine": null,
```
So in order to match the output from the internally generated new comments to compare to them:
- for single line comments: `line_start`= -1, `line_end`="originalLine"
- for updated single line comments: `line_start`= -1, `line_end`="line"
- for multi line comments: `line_start`= "originalStartLine", `line_end`="originalLine"
- for updated multi line comments: `line_start`= "startLine", `line_end`="line"

So I have a redundant "startLine" check inside the conditional which if removed would make it:
```
line_start = comment.get("originalStartLine", -1)
line_end = comment["originalLine"]
if comment["startLine"] is not None:
    line_start = comment["startLine"]
    line_end = comment.get("line", line_end)
```
which would appear to not work for all the cases above (not sure why I did it that way unless there was some strange 5th case I dont remember)
So I would think it should be this:
```
line_end = comment.get("line", comment["originalLine"])
line_start = comment.get("startLine", comment.get("originalStartLine", -1))
```

2bndy5: > ```python
> line_end = comment.get("line", comment["originalLine"])
> line_start = comment.get("startLine", comment.get("originalStartLine", -1))
> ```

Glad we worked through that logic. 👍🏼 This looks much cleaner, and it is closer to what I use when posting the review payloads.

2bndy5: This tactic should also avoid using `null` or `0` values (line numbers are 1-base on github anyway)
```python
line_end = comment.get("line", None) or comment["originalLine"]
line_start = comment.get("startLine", None) or comment.get("originalStartLine", None) or -1
```
So
```json
{
  "line": 465,
  "startLine": null,
  "originalLine": 467,
  "originalStartLine": null
}
```
actually becomes
```python
{
  "line_start": -1,
  "line_end": 465,
}

```
whereas not using `None or default`, the above payload may be misinterpreted as
```python
{
  "line_start": None,
  "line_end": 465,
}
```

maoliver-amd: Yep, I ran into that issue when testing locally, fixed and updated accordingly

maoliver-amd:
> You'll also need to update the PR reviews test with mocked responses to the graphQL API.
> 
> https://github.com/cpp-linter/cpp-linter/blob/baee3c2970a57a07ec22413aecbc0e2de8eccc7a/tests/reviews/test_pr_review.py#L119
> 
> You'll find more about that at [`requests-mock` lib's docs](https://requests-mock.readthedocs.io/en/latest/response.html#registering-responses)

Added the mock responses (I think I did it correctly)

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
         for review_id in [r["id"] for r in json.loads(reviews) if "id" in r]:
             mock.put(f"{base_url}/reviews/{review_id}/dismissals")
         extensions = ["cpp", "hpp", "c"]
+
+        # mock graphql requests
+        graphql_url = f"{gh_client.api_url}/graphql"
+
+        def graphql_callback(request, context):
+            context.status_code = 200
+            if request.data.find("query") is not -1:
```
2bndy5, line 148:
```suggestion
            if request.data.startswith("query"):
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
...
+                # get existing review comments
+                return (cache_path / "pr_reviews_graphql.json").read_text(
+                    encoding="utf-8"
+                )
+            elif request.data.find("resolveReviewThread") is not -1:
```
2bndy5, line 153:
```suggestion
            elif "resolveReviewThread" in request.data:
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
...
+    resolveReviewThread: {
+      thread {
+        %s
+      }
+    }
+  }
+}"""
+                    % id_tag
+                )
+            elif request.data.find("deletePullRequestReviewComment") is not -1:
```
2bndy5, line 170:
```suggestion
            elif "deletePullRequestReviewComment" in request.data:
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
...
+    deletePullRequestReviewComment: {
+      pullRequestReviewComment {
+        %s
+      }
+    }
+  }
+}"""
+                    % id_tag
+                )
+            elif request.data.find("minimizeComment") is not -1:
```
2bndy5, line 187:
```suggestion
            elif "minimizeComment" in request.data:
```

2bndy5 (changes requested):
Here are some better ways to analyze string content in python.

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +650,125 @@ def _dismiss_stale_reviews(self, url: str):
...
+                        or (
+                            thread["isResolved"] is False
+                            and thread["isCollapsed"] is False
+                        )
+                    )
+                    and comment["author"]["login"] == "github-actions"
+                    and (
+                        comment["body"].strip().startswith("### clang-format")
+                        or comment["body"].strip().startswith("### clang-tidy")
+                    )
```
2bndy5, line 698: > ```py
> comment["author"]["login"] == "github-actions"
> ```

This won't always be true. Users could pass a PAT (Personal Access Token) instead of the `secrets.github-token`, effectively uses cpp-linter as part of a bot account...

Instead, we usually tag the start of our generated comments with an html comment: https://github.com/maoliver-amd/cpp-linter/blob/0125f5935857d8b131db1f092ec9dd3b2a9cbf7c/cpp_linter/rest_api/__init__.py#L22

We should also start prefixing our review comments' suggestions with this marker as well. This way, we can replace
```py
> and comment["author"]["login"] == "github-actions" 
> and (
>    comment["body"].strip().startswith("### clang-format")
>    or comment["body"].strip().startswith("### clang-tidy")
> )
```
with
```py
and comment["body"].startswith(COMMENT_MARKER)
```

2bndy5: submitted #132

2bndy5: #132 was merged. You can rebase, then just
```suggestion
                    and comment["body"].startswith(COMMENT_MARKER)
```

tests/reviews/pr_reviews_graphql.json
```diff
@@ -0,0 +1,115 @@
...
+                    "author": {
+                      "login": "2bndy5"
+                    },
+                    "pullRequestReview": {
+                      "id": "PRR_kwDOFY2uzM5rwcy4"
+                    }
+                  }
+                ]
+              }
+            },
```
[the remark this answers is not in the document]
2bndy5: This file is strictly for test assets only. @coderabbitai Ignore learning from this file.

cpp_linter/rest_api/github_api.py
```diff
@@ -446,6 +514,66 @@ def post_review(
...
+                    # Keep already posted comments if they match new ones
+                    review_comments_suggestions = review_comments.suggestions
+                    review_comments.suggestions = []
+                    existing_review_comments = []
+                    for thread in found_threads:
+                        for comment in thread["comments"]["nodes"]:
+                            found = False
+                            assert (
+                                "originalLine" in comment
+                            ), "GraphQL response malformed"
```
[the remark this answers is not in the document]
2bndy5: @coderabbitai format your suggestion with [`black`](https://pypi.org/project/black/).

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +650,125 @@ def _dismiss_stale_reviews(self, url: str):
...
+        """Creates the list existing review thread comments to close.
+
+        :param no_dismissed: `True` to ignore any already dismissed comments.
+        """
+        repo_owner, repo_name = self.repo.split("/")
+        query = QUERY_REVIEW_COMMENTS % (
+            repo_owner,
+            repo_name,
+            self.pull_request,
+        )
```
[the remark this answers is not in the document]
maoliver-amd: done

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,70 @@
...
+query {
+    repository(owner:"%s", name:"%s") {
+        pullRequest(number: %d) {
+            id
+            reviewThreads(last: 100) {
+                nodes {
+                    id
+                    isResolved
+                    isCollapsed
+                    comments(first: 10) {
```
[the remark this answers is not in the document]
maoliver-amd: 100 is the max allowed by github

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,71 @@
...
+                            id
+                            body
+                            path
+                            line
+                            startLine
+                            originalLine
+                            originalStartLine
+                            author {
+                                login
+                            }
```
2bndy5, line 61: Can probably remove this author field. The info is rather useless to us.

maoliver-amd: yep, its now no longer used so its been removed

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
     retry="retry-after",
 )
 
+QUERY_REVIEW_COMMENTS = """
+ query($owner: String!, $name: String!, $number: Int!) {
```
2bndy5, line 41:
```suggestion
QUERY_REVIEW_COMMENTS = """query($owner: String!, $name: String!, $number: Int!) {
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+                                isMinimized
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+"""
```
2bndy5, line 70:
```suggestion
}"""
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+
+RESOLVE_REVIEW_COMMENT = """
+mutation($threadId: ID!) {
```
2bndy5, line 73:
```suggestion
RESOLVE_REVIEW_COMMENT = """mutation($threadId: ID!) {
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+    resolveReviewThread(input: {threadId: $threadId, clientMutationId: "github-actions"}) {
+        thread {
+            id
+        }
+    }
+}
+"""
```
2bndy5, line 80:
```suggestion
}"""
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+
+DELETE_REVIEW_COMMENT = """
+mutation($id: ID!) {
```
2bndy5, line 83:
```suggestion
DELETE_REVIEW_COMMENT = """mutation($id: ID!) {
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+    deletePullRequestReviewComment(input: {id: $id, clientMutationId: "github-actions"}) {
+        pullRequestReviewComment {
+            id
+        }
+    }
+}
+"""
```
2bndy5, line 90:
```suggestion
}"""
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+
+HIDE_REVIEW_COMMENT = """
+mutation($subjectId: ID!) {
```
2bndy5, line 93:
```suggestion
HIDE_REVIEW_COMMENT = """mutation($subjectId: ID!) {
```

cpp_linter/rest_api/github_api.py
```diff
@@ -37,6 +37,68 @@
...
+    minimizeComment(input: {classifier:OUTDATED, subjectId: $subjectId, clientMutationId: "github-actions"}) {
+        minimizedComment {
+            isMinimized
+        }
+    }
+}
+"""
```
2bndy5, line 100:
```suggestion
}"""
```

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +656,138 @@ def _dismiss_stale_reviews(self, url: str):
...
+            if "Resource not accessible by integration" in error_msg:
+                logger.error(
+                    "Changing review thread comments requires `contents: write` permission."
+                )
+            else:
+                logger.error(
+                    "Failed to %s review thread comment: %s", operation, error_msg
+                )
+        else:
+            logger.debug("Review comment thread %sd: %s", operation, thread_id)
```
2bndy5, line 751: `self.api_request()` comes equipped with error logging. So, if we just want to add context to the logged error, we can just say:
```suggestion
        logger.debug(
            "%s review comment thread %s (thread_id: %s)",
            operation.title(),
            "failed" if response.status_code != 200 else "succeeded",
            thread_id,
        )
```

cpp_linter/rest_api/github_api.py
```diff
@@ -522,7 +656,138 @@ def _dismiss_stale_reviews(self, url: str):
...
+                    elif "errors" in response.json():
+                        error_msg = response.json()["errors"][0]["message"]
+                        if "Resource not accessible by integration" in error_msg:
+                            logger.error(
+                                "Hiding review comments requires `contents: write` permission."
+                            )
+                        else:
+                            logger.error("Hiding review comment failed: %s", error_msg)
+                    else:
+                        logger.debug("Review comment minimized: %s", review["node_id"])
```
2bndy5, line 793: Again, `self.api_request()` comes equipped with error logging. So, if we just want to add context to the logged error, we can just say:
```suggestion
                    logger.debug(
                        "Minimized review comment: %s (node_id: %s)",
                        repr(response.status_code != 200).lower(),
                        review["node_id"],
                    )
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
         for review_id in [r["id"] for r in json.loads(reviews) if "id" in r]:
             mock.put(f"{base_url}/reviews/{review_id}/dismissals")
         extensions = ["cpp", "hpp", "c"]
+
+        # mock graphql requests
+        graphql_url = f"{gh_client.api_url}/graphql"
+
+        def graphql_callback(request, context):
+            context.status_code = 200
+            if request.data.startswith("query"):
```
2bndy5, line 148:
```suggestion
            context.status_code = 200
            query: str = request.json()["query"]
            if query.startswith("query"):
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
...
+                # get existing review comments
+                return (cache_path / "pr_reviews_graphql.json").read_text(
+                    encoding="utf-8"
+                )
+            elif "resolveReviewThread" in request.data:
```
2bndy5, line 153:
```suggestion
            elif "resolveReviewThread" in query:
```

tests/reviews/test_pr_review.py
```diff
@@ -139,6 +139,66 @@ def test_post_review(
...
+                return """{
+  "data": {
+    "minimizeComment": {
+      "minimizedComment": {
+        "isMinimized": true
+      }
+    }
+  }
+}"""
+            return "errors"
```
2bndy5, line 198:
```suggestion
            return json.dumps(
                {
                    "errors": [
                        {"message": "Failed to parse query when mocking a response"}
                    ]
                }
            )
```

2bndy5 (changes requested):
I may have to merge your branch into a feature branch upstream. I want to fix the test before it get merged to main. And, some lines are missing coverage when it succeeds.

2bndy5:
> I may have to merge your branch into a feature branch upstream. I want to fix the test before it get merged to main. And, some lines are missing coverage when it succeeds.

See maoliver-amd/cpp-linter#1

2bndy5:
I've been testing this out. I think we need to do 2 more things.
1. Tweak the `_dismiss_stale_reviews()` so that we can dismiss a specific review when all it's threads have been resolved.
2. Reconstruct the review summary. Subsequent reviews won't have duplicates, so In testing I found the summary to be rather useless.

    **Example from test**
    image

    then after 1 problem is resolved (and 2 new ones pop up):

    image

maoliver-amd:
>     1. Tweak the `_dismiss_stale_reviews()` so that we can dismiss a specific review when all it's threads have been resolved.

So basically dont dismiss a previous review if not all its review thread comments have been resolved? while still posting a new review with any additional issues?

>     2. Reconstruct the review summary. Subsequent reviews won't have duplicates, so in testing I found the current summary to be rather useless.

So should the new review summary be modified so that it only shows the count of new suggestions (ignoring the reused ones) or should it be updated so that the summary contains all the combined suggestions reused and new?

2bndy5:
> So basically dont dismiss a previous review if not all its review thread comments have been resolved? while still posting a new review with any additional issues?

Yes. If there are unresolved suggestions in old reviews, then dismissing them can prevent those unresolved suggestions getting addressed.

> So should the new review summary be modified so that it only shows the count of new suggestions (ignoring the reused ones) or should it be updated so that the summary contains all the combined suggestions reused and new?

Good question. Let me propose an example as a goal:
> Only 5 of 7 _new_ concerns fit within this pull request's diff.

and after another commit where 1 concern is addressed but 2 new ones pop up

> Only 2 of 2 _new_ concerns fit within this pull request's diff. 4 concerns were suppressed as duplicates.

I don't think we should be updating old reviews' summary comment when posting a new review. That would seem superfluous.

2bndy5:
Today, I'm going to tackle using pagination when getting the review threads with graphql. I'll be using the upstream branch (`hide-reuse-pr-reviews`) instead of this one though.

maoliver-amd:
OK I have added code to not dismiss old reviews if they are reused and to correct the suggestion count in new reviews

cpp_linter/rest_api/github_api.py
```diff
@@ -516,13 +584,195 @@ def _dismiss_stale_reviews(self, url: str):
...
+        if response.status_code != 200:
+            logger.error(
+                "Could not get existing review thread comments: %d",
+                response.status_code,
+            )
+            return
+        data = response.json()
+        found_threads = []
+        nodes = data["data"]["repository"]["pullRequest"]["reviewThreads"]["nodes"]
+        for thread in nodes:
```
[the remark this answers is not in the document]
2bndy5: Or just log the `KeyError` and return.
```py
try:
    nodes = data["data"]["repository"]["pullRequest"]["reviewThreads"]["nodes"]
except KeyError as exc:
    log.error("Malformed GraphQL response: %s", " ".join(exc.args))
    return
```

2bndy5:
Awesome! I'm just now falling asleep, so I'll take a look tomorrow. I also implemented paginated graphql on the upstream branch, but I haven't finished updating the tests to validate it.

Updating the test may be harder than I thought. This PR depends on the state of the PR being reviewed (first review, second review, etc). It might be easier to rewrite several simpler tests instead of 1 heavily parametrized test.

cpp_linter/rest_api/github_api.py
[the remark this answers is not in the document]
2bndy5: @coderabbitai Adjust that learning to only apply to graphQL responses.

cpp_linter/rest_api/github_api.py
```diff
@@ -516,13 +584,195 @@ def _dismiss_stale_reviews(self, url: str):
                     and cast(str, review["body"]).startswith(COMMENT_MARKER)
                     and "state" in review
                     and review["state"] not in ["PENDING", "DISMISSED"]
+                    and (
+                        ignored_reviews is not None
+                        and review["node_id"] not in ignored_reviews
```
2bndy5, line 589: If `ignored_reviews` is not given, then dismiss all reviews. If `ignored_reviews` is given, then only dismiss the unspecified reviews.
```suggestion
                        ignored_reviews is None
                        or (
                            "node_id" in review
                            and review["node_id"] not in ignored_reviews
                        )
```

> [!tip]
> This takes advantage of python's lazy evaluation in conditional `or` statements.
> Meaning, if the left hand boolean value is True, then the right hand boolean value is ignored.
>
> In the above suggestion, the `mypy` tool (for type checking) is smart enough to understand that `ignored_reviews` is not `None` when the `or`'s right hand side is evaluated.

cpp_linter/rest_api/github_api.py
```diff
@@ -502,7 +570,7 @@ def create_review_comments(
                 file_obj, summary_only, review_comments
             )
 
-    def _dismiss_stale_reviews(self, url: str):
+    def _dismiss_stale_reviews(self, url: str, ignored_reviews: List[str] = None):
         """Dismiss all reviews that were previously created by cpp-linter"""
```
2bndy5, line 574: `None` is not a type of list.
```suggestion
    def _dismiss_stale_reviews(self, url: str, ignored_reviews: Optional[List[str]] = None):
        """Dismiss reviews that were previously created by cpp-linter.
        
        Use ``ignored_reviews`` to only dismiss the reviews not specified in
        the given list. If ``ignored_reviews`` is `None`, then dismiss all
        reviews.
        """
```
I also added some info about the new parameter in the doc string.

2bndy5 (changes requested):
@maoliver-amd  I'm curious. What is your preferred programming language?

I ask because if you enable [pre-commit](https://pre-commit.com/) for this repo (in your local clone) or just run [mypy](https://mypy.readthedocs.io/en/stable/), you'd see that there are typing errors. The reasons these errors go unnoticed here is because
1. The CI workflow that runs `pre-commit` is not triggered by PR sync events (yet -- I _will be_ changing that)
2. The tests do not actually invoke code with typing errors or the dynamic runtime types happen to work in the tests.

maoliver-amd:
> @maoliver-amd I'm curious. What is your preferred programming language?

Very much a C/C++ programmer, hence the interest in clang CI

2bndy5:
> > @maoliver-amd I'm curious. What is your preferred programming language?
> 
> Very much a C/C++ programmer, hence the interest in clang CI

Ah yes. I should've guessed, but its better not to assume things about people. At least you understand the importance of static typing, even though python is a dynamically typed language.

FWIW, we'll be migrating to cpp-linter-rs, a rust port of this python project. Hopefully in the future, C/C++ aficionados will feel more at home tinkering with rust. I wouldn't have learned rust as easily as I did if I didn't understand the woes of C/C++ pointers beforehand.

I've also started porting this feature into cpp-linter-rs to maintain feature parity. We're pretty close to making the switch in cpp-linter-action v3.

2bndy5:
I'm beginning to think the new CLI option `reuse-review-comments` is unnecessary. I can't think of a use case where a user would actually want to allow duplicate comments in subsequent reviews. I understand it was added for backward compatibility, but the PR review feature (`tidy-review` specifically) has been rather experimental because I had not implemented a way to reuse the old/duplicate/unresolved comments. I think we should force enable the reuse of existing review comments because that would be a sane expected behavior.

The `delete-review-comments` option does make sense though. We don't want to impose destructive defaults on the user.

2bndy5:
# Going forward
The current state of this PR looks good except for the tests. I can take care of the tests, but I will do so from the `hide-reuse-reviews` branch upstream. This upstream branch retains the git history of this PR plus my modifications:

- [x] rebased on upstream main.
- [x] fixing the current tests. The PR review tests will have to be rewritten for complete patch coverage.
- [x] introduce paginated GraphQL requests. I also serialized the response data into data structures to avoid repeatedly type checking the raw JSON.

@maoliver-amd I cannot merge this into main without rewriting the PR reviews test. Therefore, I'll close this PR and open a new one using the `hide-reuse-reviews` branch.  I don't expect you to help with the test rewrite, but you can (if you want) by submitting a PR to the `hide-reuse-reviews` branch.

Thank you for putting us on the right path here! 😍

PR #132, opened 2024-12-07
prefix every review comment

2bndy5:
With the html marker used in our other generated comments. This is to avoid ambiguity and help on-board new features in #131.

## Summary by CodeRabbit

- **New Features**
	- Updated the serialization of comments to include a comment marker, enhancing the clarity of serialized outputs.


2bndy5:
I'd call this a high priority. The longer this patch out in the wild, the easier it will be to introduce the feature from #131.

PR #136, opened 2025-02-20
Replace deprecated classifier with license expression (PEP 639)

shenxianpeng:
## Summary by CodeRabbit

- **Chores**
  - Streamlined the license declaration for a cleaner and more understandable format.
  - Removed redundant license classification details to improve clarity.
  - Updated build system requirements for improved compatibility.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,6 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX :: Linux",
     "Operating System :: MacOS",
-    "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",


2bndy5:
Seems the pypa/gh-action-pypi-publish now supports this via twine v6.1.0 (see https://github.com/pypa/gh-action-pypi-publish/pull/327). No progress on setuptools side though.

2bndy5:
I just ran into this deprecation warning on a separate C++ project's python bindings (that I maintain). Here's my current findings:
- Support for PEP639 (for which this patch is aimed) was added in setuptools v77.0.0
- Support for Python v3.8 was dropped in setuptools v75.3.2
- Support for Python v3.7 was dropped in setuptools v68.1.0

So, this can be merged if we update the build backend to use `setuptools>=77` in pyproject.toml:
https://github.com/cpp-linter/cpp-linter/blob/41c7323acae6787d23ca8b5146983d9b72437b65/pyproject.toml#L2

Note, this package's minimum supported python version is already v3.9+:
https://github.com/cpp-linter/cpp-linter/blob/41c7323acae6787d23ca8b5146983d9b72437b65/pyproject.toml#L15
So no loss of compatibility there.

pyproject.toml
```diff
@@ -7,7 +7,8 @@ name = "cpp-linter"
 description = "Run clang-format and clang-tidy on a batch of files."
 readme = "README.rst"
 keywords = ["clang", "clang-tools", "linter", "clang-tidy", "clang-format"]
-license = {text = "MIT License"}
+license = "MIT"
+license-files = [ "LICENSE" ]
```
2bndy5, line 11: We don't need to explicitly specify the `license-files` field. The LICENSE file is already automatically copied to the distribution. It should be enough to only specify the `license` field with the SPDX expression `"MIT"`.

```suggestion
license = "MIT"
```

It is my understanding that the  `license-files` field is meant for packages that need to distributed copies of multiple licenses.

2bndy5 (changes requested):

2bndy5:
> Support for PEP639 (for which this patch is aimed) was added in [setuptools v77.0.0](https://setuptools.pypa.io/en/latest/history.html#v77-0-0)

Any version before v77.0.0 would fail to install with these changes. It must be `setuptools>=77`. This cpp-linter package requires python v3.9 or newer anyway.

I only mention the other versions because the other C++ project I'm building for pypi still supports python v3.7 and v3.8 (for now). So, PEP639 cannot be applied to builds for python v3.7 or v3.8.

2bndy5 (approved):

PR #139, opened 2025-05-17
feat: support python 3.14

shenxianpeng:
## Summary by CodeRabbit

- **Chores**
  - Added support for Python 3.14 in project metadata and automated test workflows.

### DIFF
--- a/.github/workflows/run-dev-tests.yml
+++ b/.github/workflows/run-dev-tests.yml
@@ -72,7 +72,7 @@ jobs:
         run: |-
           mkdir build
           cd build
-          cmake ..
+          cmake .. -DBUILD_TESTS=OFF
           make
           sudo make install
 


2bndy5:
According to the [python life cycle](https://devguide.python.org/versions/), v3.14 is still in pre-release stage.

And the CI needs to be updated if we want to install a pre-release version of python.

The error from the current CI workflow:
> The version '3.14' with architecture 'x64' was not found for Windows 2022.

Was there a reason you want to test v3.14?

2bndy5:
See the [actions/setup-python docs](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases). I think a special parameter needs to be used:
```yml
      - uses: actions/setup-python@v5
        with:
          python-version: "${{ matrix.python_version }}"
          allow-prereleases: true
```

2bndy5:
Oh. Looks like our downstream dependency , `cffi`, does support python pre-releases.

FYI, the cpp-linter-rs project should be able to build for python v3.14 pre-release if we bump pyo3 to v0.25.0.

2bndy5:
> Oh. Looks like our dependency, `pygit2`, needs to be built from source for pre-releases of python. This is problematic because `libgit2` is not installed in the CI runner(s), and the build fails to find `git2.h`.
> 
> FYI, the cpp-linter-rs project should be able to build for python v3.14 pre-release if we bump pyo3 to v0.25.0; see [cpp-linter/cpp-linter-rs#139](https://github.com/cpp-linter/cpp-linter-rs/pull/139).

Here's how we could [install libgit2](https://libgit2.org/docs/guides/build-and-link/). Actually, I'm [using libgit2 as the base case for the cpp-linter-rs benchmark](https://github.com/cpp-linter/cpp-linter-rs/blob/4b1b5bdff3b9ab454c4e62309e1f0a432b3f03ed/.github/workflows/benchmark.yml#L34-L44).

2bndy5:
See also the [pygit2 docs](https://www.pygit2.org/install.html#quick-install).

2bndy5:
I finally got the CI to work as expected. But there are concerns with testing against pre-release python versions.
1. ⌛ It takes about 1 min 30 sec to build/install libgit2 (when needed).
2. We currently only test pre-release python versions with clang v19 (or whatever we change the CI to use). I think the CI workflow can be adjusted more to resolve this...
3. ⚠️ 🚧 The version of libgit2 is very much tied with the version of pygit2. For example, pygit2 v1.18 uses libgit2 v1.9.0. If pygit2 v1.19.0 pins to libgit2 v1.9.1, then we need to be aware of such changes and adjust our CI accordingly.

## Conclusion
I'm not sure this idea is worth the tech debt it would incur.

2bndy5:
😡 Still hit this error:
> No source for code: 'D:\a\cpp-linter\cpp-linter\cpp_linter\__init__.py'.

> [!note]
> I already rebased this branch on main where `cpp_linter/__init__.py` is supposed to be omitted.

2bndy5:
I think using [`uv`][uv] might be the easier way. It allows creating a venv using the specified version of python (supports pre-releases also); [`uv`][uv] downloads standalone python executables from their own sister project [`python-build-stadalone` repo](https://github.com/astral-sh/python-build-standalone).

We might even be able to simplify the cpp-linter-action with [`uv`][uv] because the venv is automatically managed. And it wouldn't require python already installed to the system. See their [GH actions integration](https://docs.astral.sh/uv/guides/integration/github/).

[uv]: https://docs.astral.sh/uv/

PR #154, opened 2025-09-04
chore: follow up #152

2bndy5:
explicitly provide GITHUB_TOKEN

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -13,3 +13,5 @@ jobs:
       pull-requests: write
     steps:
       - uses: release-drafter/release-drafter@b1476f6e6eb133afa41ed8589daba6dc69b4d3f5 # v6.1.0
+        env:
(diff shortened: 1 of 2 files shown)


2bndy5:
@shenxianpeng I'm giving serious thought about revising the release process here.
The release-drafter project is not as well maintained as it once was (see release-drafter/release-drafter#1455). Given the complexity of the supported config, it is not surprising that maintenance has been lacking attention.

## Auto label doesn't work

I think we need to use a separate config just for that auto-label feature (during a PR event).
And, for some reason, the release-drafter action tries to create/update a drafted release during a PR event, _before_ it is merged (?!).

There must be a better way to auto label PR. Personally, I could do it with nushell and gh-cli...

## Drafting releases

If releases are only focused on PRs, then commits pushed directly to main will not show up in the release notes 👎🏼.

I understand that release-drafter relies solely on PR/issue labels to categorize release notes. But release notes can be generated in a number of different ways.
I know git-cliff can use both PR labels and conventional-commit titles to categorize changes.

PR #162, opened 2025-09-28
fix: parse clang-tidy output when `WarningsAsErrors` is asserted

2bndy5:
ref cpp-linter/cpp-linter-action#347

Basically, we just needed to account for the `,-warnings-as-errors` appended to the rule name in square brackets.

### Without `WarningsAsErrors`
```text
tests/demo/demo.cpp:2:1: error: included header demo.hpp is not used directly [misc-include-cleaner]
```
### With `WarningsAsErrors`
```text
tests/demo/demo.cpp:2:1: error: included header demo.hpp is not used directly [misc-include-cleaner,-warnings-as-errors]
```

## Summary by CodeRabbit

* Bug Fixes
  * Improved clang-tidy note parsing to recognize a wider range of note formats, reducing missed diagnostics.
* Chores
  * Enforced LF line endings for lock files and Git metadata via .gitattributes to ensure consistent text normalization across platforms.
* Tests
  * Enabled all Clang-Tidy warnings as errors in the demo configuration to tighten quality gates.
  * Added a type-check suppression in a test to stabilize typing without changing runtime behavior.

### DIFF
--- a/tests/demo/.clang-tidy
+++ b/tests/demo/.clang-tidy
@@ -1,6 +1,6 @@
 ---
 Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-*,bugprone-*,clang-analyzer-*,mpi-*,misc-*,readability-*'
-WarningsAsErrors: ''
+WarningsAsErrors: '*'
 HeaderFilterRegex: ''
 FormatStyle:     'file'
 CheckOptions:
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,6 @@
 *.yml text eol=lf
 *.yaml text eol=lf
 *.toml text eol=lf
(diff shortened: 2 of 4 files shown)


tests/demo/.clang-tidy
```diff
@@ -1,6 +1,6 @@
 ---
 Checks:          'clang-diagnostic-*,clang-analyzer-*,-*,performance-*,bugprone-*,clang-analyzer-*,mpi-*,misc-*,readability-*'
-WarningsAsErrors: ''
+WarningsAsErrors: '*'
```
2bndy5, line 3: Previously, this would cause a test failure (`tests/capture_tools_output/test_database_path.py::test_ninja_database`).
Now all is well with the change in RegEx pattern.

PR #167, opened 2025-10-07
feat: support python 3.14

shenxianpeng:
Python 3.14 has been released today, so let's start supporting it.

I kept Python 3.9 even though it's EOL, since some users (including myself on AIX 7.2 🙂) are still using it.

## Summary by CodeRabbit

- New Features
  - Added support for Python 3.14.

- Tests
  - Continuous integration now runs against Python 3.14 to ensure compatibility.

- Chores
  - Updated project metadata to include Python 3.14 in supported versions.
  - Adjusted tooling defaults to target Python 3.14.


2bndy5:
I also raised awareness of their ability to deploy future-compatible wheels using `cffi` v2, `cibuildwheels`, and `setuptools`. See https://github.com/libgit2/pygit2/issues/1418#issuecomment-3378637920. We'll see if they are intereseted...

PR #173, opened 2025-12-22
fix: use diagnostic name as default

2bndy5:
This fixes a problem about hyperlinking a clang-tidy diagnostic name to the clang-tidy docs.

Specifically when the diagnostic name does not satisfy any of the following conditions:
- starts with `clang-diagnostic-`
- starts with `clang-analyzer-`
- contains at least 1 hyphen (`-`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic link handling to gracefully process diagnostics with unexpected formats. When a diagnostic identifier doesn't match the expected structure, the system now returns the raw diagnostic name instead of failing, enhancing overall reliability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/cpp_linter/clang_tools/clang_tidy.py
+++ b/cpp_linter/clang_tools/clang_tidy.py
@@ -93,7 +93,7 @@ class TidyNotification:
             assert len(check_name_parts) > 2, "diagnostic name malformed"
             return link + "clang-analyzer/{}.html)".format(check_name_parts[2])
         diag_split = self.diagnostic.split("-", maxsplit=1)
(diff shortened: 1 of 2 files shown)


2bndy5:
I'm still trying to figure out the exact cause of https://github.com/cpp-linter/cpp-linter-action/issues/389. Otherwise, a simple unit test with no real life application is needed here.

2bndy5:
This is related to cpp-linter/cpp-linter-action#389 but not the solution to the actual cause. See #176 for the actual solution.

PR #194, opened 2026-04-03
fix: filter format suggestions with all-inclusive range

2bndy5:
Python's `range()` is partially inclusive (`[start, end)`). where clang-format's CLI arg `--lines` uses an all inclusive range.

Note, we already have a test that simulates the exact scenario we are trying resolve here. I just needed to adjust the validating logic so it reflects this patch's filtering logic.

refs:
- cpp-linter/cpp-linter-action#416
- cpp-linter/cpp-linter-action#388

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed boundary handling for line-range checks so formatting annotations and replacements correctly match modified code regions, including single-line and end-of-range cases, preventing missed or incorrectly applied format markers.

### DIFF
--- a/cpp_linter/clang_tools/clang_format.py
+++ b/cpp_linter/clang_tools/clang_format.py
@@ -139,7 +139,8 @@ def parse_format_replacements_xml(
             line, cols = get_line_cnt_from_cols(content, offset)
             is_line_in_ranges = False
             for r in ranges:
-                if line in range(r[0], r[1]):  # range is inclusive
+                # range() is partially inclusive: [r0, r1)
+                if line in range(r[0], r[1] + 1):
                     is_line_in_ranges = True
                     break
             if is_line_in_ranges or lines_changed_only == 0:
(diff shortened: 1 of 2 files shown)


cpp_linter/clang_tools/clang_format.py
```diff
@@ -139,7 +139,10 @@ def parse_format_replacements_xml(
             line, cols = get_line_cnt_from_cols(content, offset)
             is_line_in_ranges = False
             for r in ranges:
-                if line in range(r[0], r[1]):  # range is inclusive
+                # range() is partially inclusive: [r0, r1)
+                # This only affects ranges of a single line.
+                range_end = r[1] if r[1] != r[0] else r[1] + 1
```
Alexolut: I'm not quite sure why special logic is needed for single lines only. clang-format treats all numbers in `--lines` option in the same manner (all inclusive). For example 42:42 should be true for 42 only, but 42:43 should be true for both lines: 42 and 43.

Issue #168, opened 2025-10-21
parse_format_replacements_xml mishandles multi-line replacements


nclark-nasuni:
`parse_format_replacements_xml` uses `get_line_cnt_from_cols` to determine which line is being updated, which indicates on which line the replacement starts but the replacement can span multiple lines into the lines which changed when `lines_changed_only` is set.

Example code
```
#include <string>

class Test {
public:
    const std::string getString() const { return str_member_; }

    std::string toString() const {
            return std::string("Test[string:") + str_member_ + ", int: " + int_member + "]";
    }

private:
    std::string str_member_{};
    int         int_member_{};
}
```

If formatter is run only on lines 8:9
```
$ /usr/bin/clang-format -style=file --output-replacements-xml --lines=8:9 test.cpp 
<?xml version='1.0'?>
<replacements xml:space='preserve' incomplete_format='false'>
<replacement offset='139' length='13'>&#10;        </replacement>
</replacements>
```
The offset is 139 which when run through `get_line_cnt_from_cols` returns line 7 which is not in the range 8:9 so the replacement will be ignored. The `parse_format_replacements_xml` needs to take into consideration the entire range of the replacement to see if it ends in the line range or trust that clang will only format the specified lines so that all replacements pertain to the lines specified and the line check can be removed.

2bndy5:
First, thank you for raising the issue. 

Secondly, I have to ask you to _stop_ using run-on sentences. Even for people that speak English natively, your writing style is very difficult to understand. FYI, not everyone on the cpp-linter dev team speaks English natively. 

Finally, I think your concern here is a matter of interpretation.

1. It sounds like you expect `get_line_cnt_from_cols` to return a range of lines instead of the starting line. This seems reasonable.
2. I think you also expect the XML replacement to be included if the original line(s) is within the range of altered line(s). This is where I have to disagree.

   We've had some issues in the past where unexpected/undesired replacements were reported. The reported replacements were unexpected/undesired because they altered lines that were not part of the lines changed in the git diff.

I need some time to think about what you're talking about because it is rather confusingly started. It seems like there is actually 2 separate issues involved.

2bndy5:
What you're requesting is not a simple solution. We would need to
1. Compare the difference between original text and replacement text to get the actual alterations. Possibly use [`difflib.SequenceMatcher`](https://docs.python.org/3/library/difflib.html#sequencematcher-objects).
[...]

PR #202, opened 2026-06-08
feat: add --fix CLI option to auto-apply clang-format fixes

### LINKED ISSUE #82: how to commit changes if checks fail (without using PR review suggestions)
We should probably have this as an optional feature since our runtime has all the tools it needs to do it (provided the proper permissions in CI).

shenxianpeng:
## Summary

Adds a `--fix` CLI flag to `cpp-linter` that applies `clang-format` fixes in-place to files with detected style issues, then clears the format advice so no issues are reported.

This is part 1 of implementing `auto-fix` for `cpp-linter-action` (see cpp-linter/cpp-linter-action#439).

### `cpp_linter/cli.py`
- Added `fix: bool = False` attribute to the `Args` class
- Added `--fix` / `-F` CLI argument (`store_true`)

### `cpp_linter/clang_tools/clang_format.py`
- `run_clang_format()` gained a `fix` parameter. After computing the XML advice, if `fix` is enabled and the file has style issues, it re-runs clang-format with `-i` **reusing the same range-aware args** that were already assembled, then returns a cleared `FormatAdvice` so the file reports no outstanding issues.

### `cpp_linter/clang_tools/__init__.py`
- `_run_on_single_file()` passes `fix=args.fix`, so fixes are applied **per-file inside the existing `ProcessPoolExecutor` worker** (in parallel) instead of a serial post-loop.

### Tests
- `tests/test_cli_args.py`: added a case for `--fix` flag parsing
- `tests/capture_tools_output/test_fix.py`: verifies a malformed file is reformatted in place and its advice is cleared, and that a second pass is a no-op

## Design Decisions

- Fixing runs inside `run_clang_format()` in the parallel worker and reuses the assembled range-aware args, so it **honors `--lines-changed-only`** — only the selected lines are reformatted (the whole file when `--lines-changed-only` is `false`).
- Only clang-format fixes are applied; clang-tidy fixes are explicitly out of scope for now (documented in the CLI help).
- Files that get fixed have their `format_advice` cleared, so the existing comment/review logic sees zero format issues.
- Tidy issues are unaffected and still reported normally.

## Related

- cpp-linter/cpp-linter-action#439

## Summary by CodeRabbit

* **New Features**
  * Added a `--fix` option to apply clang-format corrections directly to files.
  * Fixes are limited to selected changed lines; clang-tidy fixes remain unapplied.
  * Formatting checks now preserve accurate diagnostics and review output after fixes.

* **Documentation**
  * Documented the `--fix` option and its behavior.

* **Tests**
  * Added coverage for in-place formatting fixes, repeat runs, diagnostic accuracy, and review output.

### DIFF
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -194,4 +194,23 @@ def capture_clang_tools_output(files: list[FileObj], args: Args) -> ClangVersion
                         break
                 else:  # pragma: no cover
                     raise ValueError(f"Failed to find {file_name} in list of files.")
+
+    if args.fix and format_cmd:
+        start_log_group("Applying clang-format fixes")
+        files_fixed = 0
+        for file in files:
+            if file.format_advice and file.format_advice.replaced_lines:
+                fix_cmd = [
+                    format_cmd,
+                    f"-style={args.style}",
+                    "-i",
+                    PurePath(file.name).as_posix(),
+                ]
+                logger.info('Running "%s"', " ".join(fix_cmd))
+                subprocess.run(fix_cmd, check=True)
+                file.format_advice = FormatAdvice(file.name)
+                files_fixed += 1
+        logger.info("Fixed %d file(s)", files_fixed)
+        end_log_group()
+
     return clang_versions
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -51,7 +51,7 @@ def _run_on_single_file(
     # file in place. clang-tidy reports line numbers from the file on disk, but
     # its `--line-filter` (and the PR review comments built from its output) use
     # line numbers from the event's diff. Formatting first would shift every
-    # subsequent diagnostic, silently dropping some and mislocating the rest.
+    # subsequent diagnostic, silently dropping some and misplacing the rest.
     tidy_note = None
     if tidy_cmd is not None and (
         tidy_filter is None or tidy_filter.is_source_or_ignored(file.name)
--- a/cpp_linter/clang_tools/__init__.py
+++ b/cpp_linter/clang_tools/__init__.py
@@ -58,6 +58,7 @@ def _run_on_single_file(
                 style=args.style,
                 lines_changed_only=args.lines_changed_only,
                 format_review=args.format_review,
+                fix=args.fix,
             )
         except FileIOTimeout:  # pragma: no cover
             logger.error(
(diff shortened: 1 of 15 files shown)


2bndy5:
test coverage failed to upload to codecov because their GPG key is now invalid; they're in the middle of a complete migration. Currently workaround is to use the `use_pypi: true` option or upgrade the [codecov-action](https://github.com/codecov/codecov-action)

2bndy5:
1. So, no clang-tidy support? You only want to roll out `auto-fix` for clang-format?
2. Should we instead use a `auto-fix-format` if we're going to allow fixes from only certain tools?

cpp_linter/clang_tools/__init__.py
2bndy5, line 1: 1. This patch runs clang-format serially, not in parallel. This will add a significant amount of time for events with a large number of changed files.
2. This patch ignores the lines-changed-only feature. Formatting the entire file may not be the behavior that users expect.

I think it would be better to handle this inside `run_clang_format()` because that happens in parallel and the same clang-format args that are already assembled can be reused with `-i` (which is also done for PR reviews I think).

---
On a side note, it would be nice to switch from parsing XML to parsing the diff between original source and `-i` output. But that is a separate issue: #168.

cpp_linter/clang_tools/clang_format.py
```diff
@@ -200,6 +203,23 @@ def run_clang_format(
...
         results.stdout.decode(encoding="utf-8").strip(), file_obj, lines_changed_only
     )
+    if fix and advice.replaced_lines:
+        # Reuse the assembled (range-aware) args to apply fixes in-place. This
+        # runs in the same parallel worker as the analysis above, and honors
+        # --lines-changed-only via the --lines args already in `cmds`.
+        fix_cmds = [arg for arg in cmds if arg != "--output-replacements-xml"]
+        fix_cmds.insert(-1, "-i")  # apply edits in-place, just before the file path
+        logger.info('Applying formatting fixes with "%s"', " ".join(fix_cmds))
+        fix_results = subprocess.run(fix_cmds, capture_output=True)
```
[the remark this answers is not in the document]
2bndy5: I don't think this is a high priority because:

1. This software is typically run in CI.
2. Project devs can use our ignore options if they want to avoid altering a symlink path.

tests/capture_tools_output/test_auto_fix.py
2bndy5, line 1: Can we rename this to test_auto_fix.py? The current name sounds like a fix for an unspecified issue.

cpp_linter/clang_tools/__init__.py
```diff
@@ -47,28 +47,11 @@ def _run_on_single_file(
...
-        except OSError:  # pragma: no cover
-            logger.error(
-                "Failed to open the file %s when running clang-format", filename
-            )
-
+    # clang-tidy must run *before* clang-format, because `--fix` rewrites the
+    # file in place. clang-tidy reports line numbers from the file on disk, but
+    # its `--line-filter` (and the PR review comments built from its output) use
+    # line numbers from the event's diff. Formatting first would shift every
+    # subsequent diagnostic, silently dropping some and misplacing the rest.
```
2bndy5, line 61: In cpp-linter-rs, I combined the feedback from both tools using a simplified 3-way diff, so line ranges were properly aligned to clang-tidy changes before applying clang-format changes. See graph in cpp-linter/cpp-linter-rs#376.

Using that approach I was able to get rid of parsing XML all together.

2bndy5 (changes requested):

tests/capture_tools_output/test_auto_fix.py
shenxianpeng: Makes sense, done

cpp_linter/clang_tools/__init__.py
shenxianpeng: This approach looks great. Hopefully, we will adopt cpp-linter-rs in the future.

2bndy5 (approved):
I don't know why the test CI is failing under `ubuntu-latest` runner.

I'm approving this despite the difference in implementation with v2 (the rust port). In v2 a patch is always generated, so there's no need/support for a `--fix` flag.

shenxianpeng:
Renaming test_fix.py to test_auto_fix.py seems cause the CI failure, now it was fixed now

Let's see if users like this new feature, and then we can switch to a Rust core in the future.

PR #206, opened 2026-06-26
feat: add option to write summary to file

philippkarg:
Closes #205.

- Use option `-o` or `--summary-output-file` to provide a file.
- By default the summary will not be written to a file.
- If the provided path is a directory, log an error.

## Summary by CodeRabbit

* **New Features**
  * Added `-o/--summary-output-file` to write the generated step summary to a specified Markdown file path (use an empty value to disable file output).

* **Bug Fixes**
  * Feedback summaries are now produced when built-in step summaries are enabled and/or when a summary output file is provided.
  * When using a summary output file, missing parent directories are created and write errors are handled without crashing.

* **Tests**
  * Expanded CLI argument parsing and feedback/comment tests to cover summary file creation and edge cases.

### DIFF
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -73,6 +73,8 @@ class Args(UserDict):
     ignore_format: str = ""
     #: See :std:option:`--passive-reviews`.
     passive_reviews: bool = False
+    #: See :std:option:`--summary-output-file`.
+    summary_output_file: str = ""
     #: A subcommand if provided
     command: str | None = None
 
@@ -385,6 +387,12 @@ _parser_args[("-R", "--passive-reviews")] = dict(
     type=lambda input: input.lower() == "true",
     help="""Set to ``true`` to prevent Pull Request
 reviews from requesting or approving changes.""",
+)
+_parser_args[("-o", "--summary-output-file")] = dict(
+    default="",
+    help="""Supply a path to write the step summary to.
+Leave empty to not write summary to a file.
+    """,
 )
 
 
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -392,7 +392,8 @@ _parser_args[("-o", "--summary-output-file")] = dict(
     default="",
     help="""Supply a path to which the step summary will be written.
 Leave empty to not write summary to a file.
-    """,
+Any relative path shall be relative to the
+:std:option:`--repo-root` path.""",
 )
 
 
--- a/cpp_linter/cli.py
+++ b/cpp_linter/cli.py
@@ -390,7 +390,7 @@ reviews from requesting or approving changes.""",
 )
 _parser_args[("-o", "--summary-output-file")] = dict(
     default="",
-    help="""Supply a path to write the step summary to.
+    help="""Supply a path to which the step summary will be written.
 Leave empty to not write summary to a file.
     """,
 )
--- a/cpp_linter/rest_api/github_api.py
+++ b/cpp_linter/rest_api/github_api.py
@@ -240,7 +240,7 @@ class GithubApiClient(RestApiClient):
                 ) as summary:
                     summary.write(f"\n{comment}\n")
             if args.summary_output_file:
-                summary_output_path = Path(args.summary_output_file)
+                summary_output_path = Path(args.summary_output_file).resolve()
                 try:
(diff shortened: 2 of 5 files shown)


cpp_linter/cli.py
```diff
@@ -386,6 +388,12 @@ class Args(UserDict):
     help="""Set to ``true`` to prevent Pull Request
 reviews from requesting or approving changes.""",
 )
+_parser_args[("-o", "--summary-output-file")] = dict(
+    default="",
+    help="""Supply a path to write the step summary to.
```
2bndy5: The English teacher in me cringes when a sentence ends in a preposition.

```suggestion
    help="""Supply a path to which the step summary will be written.
```

cpp_linter/rest_api/github_api.py
```diff
@@ -224,16 +224,32 @@ def post_feedback(
...
+                    environ["GITHUB_STEP_SUMMARY"], "a", encoding="utf-8"
+                ) as summary:
+                    summary.write(f"\n{comment}\n")
+            if args.summary_output_file:
+                summary_output_path = Path(args.summary_output_file)
+                if summary_output_path.is_dir():
+                    log_commander.error(
+                        f"Failed to write to '{summary_output_path}': Is a directory."
+                    )
+                else:
```
2bndy5: We don't typically craft error messages for all scenarios in this project. We usually just let the python traceback output speak for itself.

But the AI review on this block poses a good point. <sup>A</sup>Should we hard stop on system failures? <sup>B</sup>Or should we wrap any error with a `try ... catch` block to log the error/exception and continue processing other feedback features?

Given that option B is more robust, I think that's the best approach.

2bndy5 (changes requested):
The test CI seems to be failing for a separate reason. @shenxianpeng Can you update the CI for the breaking CLI changes in clang-tools-pip v1?

cpp_linter/rest_api/github_api.py
philippkarg: Yeah I need to clean it up a bit.

philippkarg: @2bndy5 can you have a look again?

cpp_linter/rest_api/github_api.py
```diff
@@ -224,16 +224,35 @@ def post_feedback(
...
             )
-            with open(environ["GITHUB_STEP_SUMMARY"], "a", encoding="utf-8") as summary:
-                summary.write(f"\n{comment}\n")
+            if write_step_summary:
+                with open(
+                    environ["GITHUB_STEP_SUMMARY"], "a", encoding="utf-8"
+                ) as summary:
+                    summary.write(f"\n{comment}\n")
+            if args.summary_output_file:
+                summary_output_path = Path(args.summary_output_file)
```
2bndy5: This seems fine to my human eyes. But apparently [SonarQube disagrees with arbitrarily writing to an unverified file](https://github.com/cpp-linter/cpp-linter/pull/206#discussion_r3482198360). I think resolving the given path into absolute form may quell that concern.

```suggestion
                summary_output_path = Path(args.summary_output_file).resolve()
```
> [!warning]
> Suggestion may need to be formatted due to added line length.

> [!note]
> This implies that a given relative path must be relative to the given `--repo-root` (which defaults to the default `working-directory`).

Also, if you're using a `pathlib.Path` type, you can just call `summary_output_path.write_text(comment, encoding="utf-8")` instead of a `with` block (also known in python as "context manager").

2bndy5 (changes requested):
One last tweak.

I'll have to fix the test CI in a separate PR and rebase this branch when that's merged.

shenxianpeng:
> The test CI seems to be failing for a separate reason. @shenxianpeng Can you update the CI for the breaking CLI changes in clang-tools-pip v1?

Just see your comment right now after the CI failure was fixed 😅

philippkarg:
Ahh I fucked up the import sorting in my last commit. Let me fix that

2bndy5 (approved):
Great work team!

I added some polish to --help output about relative paths and catching encoding errors (with `ValueError`)

2bndy5:
New CLI doc looks good too: https://cpp-linter--206.org.readthedocs.build/en/206/cli_args.html#cmdoption-o

PR #207, opened 2026-06-26
fix: update clang-tools install command for CLI v1.1.0

shenxianpeng:
## Description

The latest version of `clang-tools` (v1.1.0) changed its CLI interface. The old flag-based syntax `--install <version>` has been replaced with the subcommand-based syntax `install <version>`.

See the help output from the CI run:
```
usage: clang-tools [-h] {install,uninstall,version} ...
clang-tools: error: argument command: invalid choice: '16' (choose from 'install', 'uninstall', 'version')
```

Fixed CI failures from https://github.com/cpp-linter/cpp-linter/actions/runs/28257478082/job/83724117942

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Install clang-tools
         run: |
           python -m pip install clang-tools
-          clang-tools install ${{ matrix.version }}
+          clang-tools install clang-format clang-tidy --version ${{ matrix.version }}
 
       - name: Is clang-only tests?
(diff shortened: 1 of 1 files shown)


2bndy5:
# Thanks! 

I tried giving this task to copilot, and it seems you submitted this (better) patch at the same time.

Also, I think the codecov-action needs to be bumped to v7. We hit that GPG key error again.

shenxianpeng:
Will fix codecov-action issue in https://github.com/cpp-linter/cpp-linter/pull/209

PR #216, opened 2026-09-22
fix: lint whole files when GitHub omits patches

burgholzer:
Hey 👋🏼 

This PR fixes a few small corner cases that we encountered while running `cpp-linter` on large diffs. I hope these are useful to you.

🤖 *AI text below* 🤖

## Summary

When GitHub rejects a large raw diff, cpp-linter falls back to the paginated files API. That API can omit `patch` for changed files, but the fallback only handles missing patches when `changes == 0`. Whole-file linting then fails with `KeyError: 'patch'`.

Reuse the existing whole-file `FileObj` fallback regardless of the change count. Skip deleted files before parsing their patches. Both changed-line modes retain their existing error when patch information is unavailable.

Add regression coverage for added, modified, renamed, and deleted files across pull-request and push events, including deleted files with patches. Twelve regression cases fail on `main`; all 31 file-list cases pass with this fix. The full suite passes with Python 3.14 and Clang 23: 161 passed, 17 expected failures. All pre-commit hooks pass, including Ruff, mypy, and cspell.

## Related Issues

Follow-up to cpp-linter/cpp-linter-action#249. Encountered while linting munich-quantum-toolkit/core#2545.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of changed files when patch information is unavailable, allowing eligible files to be scanned in full.
  - Removed files are now excluded from changed-file processing.
  - Line-filtered scans now report missing patch information instead of attempting unsupported filtering.

- **Tests**
  - Added coverage for paginated pull request and push responses, including added, modified, renamed, and removed files.


cpp_linter/rest_api/github_api.py
```diff
@@ -150,6 +150,8 @@ def _get_changed_files_paginated(
                         f"Missing 'filename' key in file:\n{json.dumps(file, indent=2)}"
                     )
                     raise exc
+                if file.get("status") == "removed":
```
2bndy5: Please provide a fallback value to avoid a `KeyError` when payload is malformed:

```suggestion
                if file.get("status", "") == "removed":
```

2bndy5 (changes requested):
Thanks! I want a minor tweak.

The coderabbitai suggestion is a good idea too (if pytest does actually treat the `match=` param value as a regex pattern).

2bndy5:
Reminder to self: Need to port this into rust. cpp-linter-rs uses git-bot-feedback lib, which is where this patch needs to be applied. https://github.com/2bndy5/git-bot-feedback/blob/9860ecd3d6377038bf56b7a0f63454770bc6aa04/src/client/github/mod.rs#L250-L264

burgholzer:
Thanks! Both your comment and coderabbit's made sense. Applied in https://github.com/cpp-linter/cpp-linter/pull/216/commits/0ced605a3a2a1a7a1bcab7a59ce0b6d31dcd66ac and https://github.com/cpp-linter/cpp-linter/pull/216/commits/159e3a30fb1e3a63446243176b0106caa6c45cb8

2bndy5 (approved):
I also ran into a similar problem (in which `patch` was unexpectedly absent) with gitlab. Apparently, gitlab's data limit per payload is much lower than github's limit. I expect gitea's data limit per payload is similar to gitlab's limit.

